PR #322, opened 2014-04-27
Pyk3 string compatibility

mariocesar:
There are two fixes:

1. Replacing `u''` with `''` to make python3.3 syntax valid, and maintain support for python2

2. Fields templates, could be the case that a field is extended from Layout and this provide full path to the template name.   For example: 'app/custom_field.html' instead of 'app/%s/custom_field.html' where the `%s` argument will be replaced with the template pack in use Django crispy forms should support both cases


maraujop:
Fix in dev branch now, see #417 thanks

Please read my changelog:
2b808e9#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR9

Please Mario, let me first sincerely apologize for not answering since 2014. If you are still interested in getting part of your PR in, please reopen a new one and I will have a look at it asap.

Cheers,
Miguel

PR #410, opened 2015-03-02
Python 2 to 3 Basestring

patgmiller:
Use six package to handle checking for base string type.


maraujop:
Thanks for your PR Patrick,

However there is no need for an external dependency here like six, because we've got a simple shim python 3 compatibility. However in one of my last patches I forgot to use it.

Fix in dev branch now, see #417 thanks

Please read my changelog:
https://github.com/maraujop/django-crispy-forms/commit/2b808e955c74c87b31b8d2d66fb101f35cf15767#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR9

Cheers,
Miguel

PR #470, opened 2015-08-18
Сonvert test_tags.py module to py.test style

Stranger6667:
Also basic fixture for template-related settings setup added. It has same functionality like [CrispyTestCase](https://github.com/maraujop/django-crispy-forms/blob/dev/crispy_forms/tests/base.py#L19)

Main changes are:
* asserts converted from TestCase's to standard python
* test classes are removed
* current_template_pack is replaced with settings.CRISPY_TEMPLATE_PACK
* the rest of the code is just indented left

### DIFF
--- a/crispy_forms/tests/conftest.py
+++ b/crispy_forms/tests/conftest.py
@@ -3,6 +3,8 @@ from django.conf import settings
 
 import pytest
 
+from .base import template_dirs, template_loaders
+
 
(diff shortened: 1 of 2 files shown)


Stranger6667:
One notice about this fixture. Seems it is not necessary, because everything is work without it. The goal was to repeat CrispyTestCase functionality. If it will not be necessary, I'll remove it.

PR #474, opened 2015-08-18
Convert test_layout module to py.test style.

Stranger6667:
Also move TEMPLATE_DIRS and TEMPLATE_LOADERS settings to test_settings module.


Stranger6667:
[There](https://github.com/maraujop/django-crispy-forms/blob/dev/crispy_forms/tests/test_layout.py#L399) was a bug:
```python
self.assertTrue(html.count('1000'), 2)
```
And proper assert should be ```assertEqual```.
In this branch this bug is fixed with conditional string check.

PR #477, opened 2015-08-18
Replace global variable with lazy object

Stranger6667:
#472 
To solve the problem with minimal code changes I've replaced TEMPLATE_PACK variable with lazy object from django.


Stranger6667:
Probably, to keep backward compatibility, class-level attributes should be lazy too.

PR #482, opened 2015-08-19
Initial approach for parametrization

Stranger6667:
Marks are rewritten with different approach, because ```settings``` variable is changing during tests, and module-level import doesn't allow to use such marks with settings parametrization.

```runtests.py``` are removed in favor of single command execution in ```Makefile```.

Unfortunately some laziness isn't lazy enough, so some parts were rewritten too.


Stranger6667:
Also, I want to notice, that with for loop in current runtests.py only coverage for latest template pack (bootstrap3) is pushed to https://codecov.io. So, after that coverage report for all template packs will be available.

PR #502, opened 2015-08-21
Remove Python 2.5 shims, unused files, imports, PEP8

Stranger6667:


pydanny:
Ahem... I remember writing those getters and setters back in the old days.

PR #520, opened 2015-09-29
add buttonholder template to bootstrap3

schrd:


flxfxp:
Hi there! Any idea when a new release will be cut that includes this template? Thanks!

PR #564, opened 2016-01-15
Include test templates in source tarball.

fladi:
Otherwise tests would fail for people who use the pypi tarball.


fladi:
Sorry, this should have been targeted at the `dev` branch.

PR #630, opened 2016-09-17
Fix missing _active_originaly_included #526.

michaelkuty:
Closes https://github.com/django-crispy-forms/django-crispy-forms/issues/526


zoidbergwill:
Looks good to me :+1:

PR #642, opened 2016-10-12
Error message class for Bootstrap 4

se210:
Bootstrap 4 uses form-control-feedback instead of help-block.


zoidbergwill:
Sounds good to me. Nice catch. :+1:

I see it mentioned in the [Forms docs](https://v4-alpha.getbootstrap.com/components/forms/#validation), but it's not in the migration guide for some reason.

PR #657, opened 2016-11-29
Support several bootstrap sizes in label_class for checkbox input

kalekseev:
For checkbox input label is displayed after the input
so the input have to be aligned with the rest of the form
using col-<device>-offset-<size> class. But sometimes form may have
different sizes on different devices so we have to build offsets
for each pair of device and size.

I'm not sure about removing label_size and bootstrap_device_type attrs but right now they are only used in the case of bootstrap checkbox.

### DIFF
--- a/crispy_forms/templatetags/crispy_forms_tags.py
+++ b/crispy_forms/templatetags/crispy_forms_tags.py
@@ -184,7 +184,6 @@ class BasicNode(template.Node):
             'flat_attrs': attrs.get('flat_attrs', ''),
             'error_text_inline': attrs.get('error_text_inline', True),
             'label_class': attrs.get('label_class', ''),
-            'label_size': attrs.get('label_size', 0),
             'field_class': attrs.get('field_class', ''),
             'include_media': attrs.get('include_media', True),
         }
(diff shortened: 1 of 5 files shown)


carltongibson:
Hi @kalekseev Could you add a before and after screenshot so I can see how this looks?

Also is there a little more we can do to test the new behaviour, or is **just** adjusting the existing texts sufficient?

crispy_forms/templatetags/crispy_forms_tags.py
```diff
@@ -184,7 +184,6 @@ def get_response_dict(self, helper, context, is_formset):
             'flat_attrs': attrs.get('flat_attrs', ''),
             'error_text_inline': attrs.get('error_text_inline', True),
             'label_class': attrs.get('label_class', ''),
-            'label_size': attrs.get('label_size', 0),
```
kalekseev: because I removed all the code where it was used in the project.

kalekseev:
with  `field_class="col-md-9 col-sm-10"` `label_class="col-md-3 col-sm-2"`

before on sm size

screenshot 1

after on sm size

screenshot

> Also is there a little more we can do to test the new behaviour, or is just adjusting the existing texts sufficient?

I think yes because we __just__ adjusted the existing behavior which was tested by these tests.

carltongibson:
OK. Thanks for the follow-up. Makes sense. I'll do a final review again today.

PR #664, opened 2016-12-14
Prevent adding names to inputs when the former are None

illia-v:
There is a small inconvenience: when a person adds a button using the class `Submit`, one is forced to give a name to it.

But often there is no need for a name. And I think, the person will choose assignment `None` as a name then.

But when one cooses the variant, a name is assigned anyway and it's `None`.

Consequently when the form is submited data which is sent to a server looks like:
`csrfmiddlewaretoken:EHE435eIo9sWz3EGdeB7ibHVz0sf3mItjD0bFfOljC07uk7oEN5ePiZ6NNFFHFKv`
`...`
`None:Submit`

I've tried to fix it by creating this pull request.

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -176,7 +176,8 @@ class BaseInput(TemplateNameMixin):
     template = "%s/layout/baseinput.html"
 
     def __init__(self, name, value, **kwargs):
-        self.name = name
+        if name is not None:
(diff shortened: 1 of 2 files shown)


illia-v:
Fixed in django/django@4a51ba228b2c416a115aee3bc034683eb02482c3

PR #672, opened 2016-12-26
Remove `utils.list_union` because it is not used

illia-v:


zoidbergwill:
Awesome. Looks good to me. :+1:

Thanks for the contribution, @illia-v.

PR #675, opened 2017-01-01
Use `django.utils.six` instead of `crispy_forms.compatibility`

illia-v:


kevin-brown:
Once Django drops support for Python 2, which if I remember correctly is coming very soon, they will stop bundling `six` with Django. Because of this, they have been recommending that third-party applications bundle their own versions of `six` now, to avoid having to bundle it later on.

illia-v:
@kevin-brown, thanks. It's convincing for me.

PR #699, opened 2017-03-20
Add for attribute in label for RadioSelect

PabloLefort:


crispy_forms/templates/bootstrap/layout/radioselect.html
```diff
@@ -5,7 +5,7 @@
     {% include 'bootstrap/layout/field_errors_block.html' %}
 
     {% for choice in field.field.choices %}
-        <label class="radio{% if inline_class %} {{ inline_class }}{% endif %}">
+        <label for="id_{{ field.html_name }}_{{ forloop.counter }}" class="radio{% if inline_class %} {{ inline_class }}{% endif %}">
```
zoidbergwill: It sounds like this isn't necessary if the `input` is within the label:

> A label element can have both a for attribute and a contained control element, as long as the for attribute points to the contained control element.

[`for` `label` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Attributes)

PabloLefort: Yes, but, in some cases when you need to hide the input to use after and/or before pseudo elements this tweak keeps onclick behaviour.
On the other hand, i'll try it with a contained input.

zoidbergwill: I'm happy to merge this then. Don't see any reasons not to.

Can you resolve the conflicts?

PabloLefort: Of course! :grinning: 

Thanks!

zoidbergwill:
LGTM :+1:

I've made the change for Bootstrap 3 and 4 as well. As soon as the tests are done, we can merge.

PR #730, opened 2017-07-28
Removed support for Python 3.2 and Django 1.9

michael-k:
Both are passed their end-of-life date; see
- https://www.python.org/dev/peps/pep-0392/
- https://www.djangoproject.com/weblog/2017/apr/04/security-releases/


zoidbergwill:
LGTM :+1:

PR #731, opened 2017-07-28
Updated suported python versions in classifiers

michael-k:


zoidbergwill:
We're testing against more version in [Travis](https://github.com/django-crispy-forms/django-crispy-forms/blob/dev/.travis.yml), but I think this is still a great improvement and more accurate. :+1:

LGTM. Thanks for your contribution.

michael-k:
The only difference is python 3.2 and that reached its EOL in February 2016: https://www.python.org/dev/peps/pep-0392/#lifespan

zoidbergwill:
Yeah, think we can happily get rid of it then too.

michael-k:
see #730

PR #735, opened 2017-08-22
Bootstrap 4 Support

zoidbergwill:
Related to #732


zoidbergwill:
Wow, running this branch on [crispy-test-project](https://github.com/django-crispy-forms/crispy-test-project) looks really ugly.

<details>
image
</details>

The horizontal/inline form stills looks good though.

I think the issues are because of the switch to flexbox.

Atrox:
Hey @zoidbergwill,

the `row` on the `form-group` divs destroys the layout. In BS4 Beta you don't want the `row` on there, only if it is a horizontal form. Currently the class is always on there (except for inline forms).

As I'm not really familiar with the codebase - the problem I encountered while fixing this locally is, how can we find out if the user wants a horizontal form and only then draw the `row`? There is no `form-horizontal` or something like that in BS4 which the user needed to specify in BS3 as a class.

Let me know if I can help on any front on this :)

zoidbergwill:
Yeah, I had some stuff in the [crispy-test-project](https://github.com/django-crispy-forms/crispy-test-project/pull/3/files) that I used to tidy them up, but I think we need a more elegant solution.

Atrox:
Pull request is out on your fork - https://github.com/zoidbergwill/django-crispy-forms/pull/1
Works as expected. The (maybe) only nitpick is, the class `form-horizontal` has to be specified even though this class no longer exists in BS4

zoidbergwill:
Awesome, thanks for your review and input @Atrox 

I just made a commit trying a similar thing: https://github.com/django-crispy-forms/django-crispy-forms/pull/735/commits/d19288d9acb2583b280bdacb9d318c04132fb82d

Except checking for `col-` on label and field classes instead. Your option might be cleaner.

zoidbergwill:
Which also has an edge case of someone using other classes that contain `col-` in them.

Atrox:
I'm not sure about the side effects this would cause because what if I don't want a horizontal form but inputs side-by-side. Wouldn't this cause problems with that method because I have to use `col-` classes on the inputs?

carltongibson:
@zoidbergwill Any chance we can get the tests passing and merge this in as is? (It's an improvement on what we have right?) 

That way we can take small incremental PRs clearing up the remaining issues.

zoidbergwill:
Sounds good, will make some time to polish this off when I get a chance this week.

zoidbergwill:
Tests pass. ✨

CesarBustios:
Hey guys! I really appreciate all the effort.

I've been working with prepend icons in the forms but they suddenly stop working:

selection_604

Do you know how can I fix it? Thanks!

carltongibson:
Hey @CesarBustios. You'll need to open a new issue with a minimal reproduce for us to be able to look at it.

CesarBustios:
Got it @carltongibson, I'll do it soon. I think it has to do with the new beta version and the CSS classes for prepends.

zoidbergwill:
@CesarBustios It will be fixed by https://github.com/django-crispy-forms/django-crispy-forms/pull/771

PR #758, opened 2017-11-24
Honor element ID set in Django for radio and multi checkboxes

nijel:
All other elements honor this correctly, but radio and multi checkboxes
do not.


zoidbergwill:
LGTM :+1:

zoidbergwill:
Thanks for the contribution!

Issue #754, opened 2017-11-04
update_attributes with css_class results in adding 'css_class' instead of 'class'


chapkovski:
* Package version: 1.7.0
* Django version: 1.11
* Python version:3.6.1
* Template pack: (Optional)

### Description:
an attempt to update the class using _update_attributes_ results in adding 'css_class' instead of 'class:

```
          self.helper.layout = Layout(
     
            Field( 'comment', css_class="form-control"),
            ButtonHolder(
                Submit('submit', Create', css_class='btn btn-success')
            )
        )
     
        self.helper['comment'].update_attributes(css_class="hello")
```

and the result is:

```
<div class="controls">
  <textarea name="comment" 
          cols="40" rows="10" 
          class="form-control textarea" 
          css_class="hello" 
          id="id_comment"> </textarea> 
</div>
```

zoidbergwill:
Thanks for filing a bug.

Just to clarify, you're doing

```
        self.helper['comment'].update_attributes(css_class="hello")
```

Should it be the following instead?

```
        self.helper['comment'].update_attributes(class="hello")
```

I'll take a look at reproducing this tomorrow.

chapkovski:
no, we can't use word class, because it is a reserved keyword in python (https://docs.python.org/3/reference/lexical_analysis.html)

zoidbergwill:
What template pack are you using? Or does this affect all template packs?

zoidbergwill:
I can reproduce it with bootstrap 4.

We use `css_class` as an example in the docs but test it with other attributes being updated.

zoidbergwill:
This PR should fix them: https://github.com/django-crispy-forms/django-crispy-forms/pull/764

PR #764, opened 2017-12-19
CSS classes are special in update_attributes

zoidbergwill:
Like they're special in Div and Field constructors

Fixes https://github.com/django-crispy-forms/django-crispy-forms/issues/754

### DIFF
--- a/crispy_forms/layout_slice.py
+++ b/crispy_forms/layout_slice.py
@@ -143,12 +143,18 @@ class LayoutSlice(object):
                 else:
                     function(layout_object)
 
-    def update_attributes(self, **kwargs):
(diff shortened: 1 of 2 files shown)


zoidbergwill:
PR tests worked but not push. Wat.

zoidbergwill:
Ahhh, failed cloning. Running again

zoidbergwill:
Oh, it's just confused cuz I pushed to origin instead of my fork then deleted it. 😞

zoidbergwill:
Maybe I need to test adding to existing classes too so coverage increases.

zoidbergwill:
Tests passing. 🎉

PR #766, opened 2018-01-15
prepended/appended text fields need row class too

zoidbergwill:
Related to https://github.com/django-crispy-forms/django-crispy-forms/issues/732#issuecomment-345230015

Missed in https://github.com/django-crispy-forms/django-crispy-forms/pull/735


zoidbergwill:
Uncovered this when updating [crispy-test-project](https://github.com/django-crispy-forms/crispy-test-project/pull/4)

Issue #781, opened 2018-03-07
Multicheckboxes HTML IDs breaking change


browniebroke:
* Package version: 1.7.1
* Django version: 1.11.11
* Python version: 3.6.4
* Template pack: (Optional) Bootstrap 3

### Description:

I've upgraded to the latest version of django-crispy-forms and our automated QA scripts started failing because it couldn't find the IDs checkboxes in a multi-select widget.

For instance, a field called ages used to generate ids like this `#id_ages_1`, `#id_ages_2`, ... but it's now `#id__1`, `#id__2`... In my case, the field uses `crispy_forms.bootstrap.InlineCheckboxes`.

This seems to be related to:
https://github.com/django-crispy-forms/django-crispy-forms/pull/758

While it's not a problem for a page a single multi-select field, it would cause non-unique IDs if a page were to have more than one field like this.

### Preferably also include:

- [x] Example Django Crispy Forms code
- [ ] Screenshots
- [x] Actual HTML generated
- [x] Expected HTML

I'll see if I can attach a more self-contained example...

browniebroke:
The `CheckboxesSampleForm` from the test may be used to reproduce the problem.

Current output:

```html
<form method="post">
  <div id="div_id_checkboxes" class="control-group">
    <label for="" class="control-label requiredField">
      Checkboxes<span class="asteriskField">*</span>
    </label>
    <div class="controls">
      <label class="checkbox">
        <input type="checkbox" checked="checked" name="checkboxes" id="id__1" value="1">Option one
      </label>
      <label class="checkbox">
        <input type="checkbox" name="checkboxes" id="id__2" value="2">Option two
      </label>
      <label class="checkbox">
        <input type="checkbox" name="checkboxes" id="id__3" value="3">Option three
      </label>
    </div>
  </div>
  <div id="div_id_alphacheckboxes" class="control-group">
    <label for="" class="control-label requiredField">
      Alphacheckboxes<span class="asteriskField">*</span>
    </label>
    <div class="controls"><label class="checkbox">
      <input type="checkbox" name="alphacheckboxes" id="id__1" value="option_one">Option one
    </label>
      <label class="checkbox">
        <input type="checkbox" checked="checked" name="alphacheckboxes" id="id__2" value="option_two">Option two
      </label>
      <label class="checkbox">
        <input type="checkbox" checked="checked" name="alphacheckboxes" id="id__3" value="option_three">Option three
      </label>
    </div>
  </div>
[...]

PR #782, opened 2018-03-07
Ambiguous IDs for multiple choice inputs

browniebroke:
- [x] Add an assertion to show the issue in #781 
- [ ] Fix?


browniebroke:
Not sure what's up with the extra commits, I'm going to re-open from a clean branch. Apologies.

Issue #790, opened 2018-03-09
"Delete" label not shown when using bootstrap4 table_inline_formset.html


simonpanay:
* Package version: 1.7 and following at least
* Django version: 1.11
* Python version: 3.5
* Template pack: bootstrap4 (possibly others)

### Description:

When using bootstrap4 "table_inline_formset.html", the "Delete" label is not displayed in the table head.
Which is confusing because "delete" checkboxes are displayed whithout any information for the end user

![capture d ecran de 2018-03-09 15-58-00](https://user-images.githubusercontent.com/124813/37213640-f13d5fca-23b2-11e8-8c6f-377300706a24.png)

smithdc1:
Fixed by #791

PR #791, opened 2018-03-09
"Delete" label should be rendered as well

simonpanay:
Is there a reason to hide Labels related to checkboxes ?


simonpanay:
This would fix #790

simonpanay:
Sounds good to me !
Thanks a lot !

Issue #793, opened 2018-03-12
InlineRadios Is Not Using form-check-inline Class Correctly


deanpienaar:
* Package version: 1.7.2
* Django version: 2.0.3
* Python version: 3.6.2
* Template pack: Bootstrap 4

### Description:
The `.form-check-inline` bootstrap class doesn't seem to be being utilised correctly. This is the current format after rendering:

```html
<div id="div_id_test_field" class="form-group">

  <label for="id_test_field_1" class="requiredField">
    Test Field<span class="asteriskField">*</span>
  </label>

  <div class="form-check">

    <label for="id_id_test_field_1" class="form-check-inline">
      <input type="radio" 
             class="form-check-input" 
             name="test_field" 
             id="id_id_test_field_1" 
            value="value1">
      Value 1
    </label> 

    <label for="id_id_test_field_2" class="form-check-inline"> 
      <input type="radio" 
             class="form-check-input" 
             name="test_field" 
             id="id_id_test_field_2" 
             value="value2">
       Value 2
    </label> 

    <label for="id_id_test_field_3" class="form-check-inline">
      <input type="radio" 
             class="form-check-input" 
             name="test_field"
             id="id_id_test_field_3" 
             value="value3">
      Value 3
    </label>

  </div>

</div>
```
![](https://puu.sh/zG1yH/72993b8f9a.png)

What would be preferable is something like

```html
<div id="div_id_test_field" class="form-group">

  <label for="id_test_field" class="requiredField">
    Test Field<span class="asteriskField">*</span>
  </label>

  <div>
    <div class="form-check form-check-inline">
      <input type="radio" 
             class="form-check-input" 
             name="test_field" 
             id="id_id_test_field_1" 
             value="value1">
      <label class="form-check-label" for="id_id_test_field_1">
        Value 1
      </label>
    </div>

    <div class="form-check form-check-inline">
      <input type="radio" 
             class="form-check-input" 
             name="test_field" 
             id="id_id_test_field_2" 
             value="value2">
      <label class="form-check-label" for="id_id_test_field_2">
        Value 2
      </label>
    </div> 

    <div class="form-check form-check-inline">
      <input type="radio" 
             class="form-check-input" 
             name="test_field" 
             id="id_id_test_field_3" 
             value="value3">
      <label class="form-check-label" for="id_id_test_field_3">
        Value 3
      </label>
    </div>
  </div>

</div>
```
[...]

PR #794, opened 2018-03-12
Fix InlineRadios Bootstrap 4 Layout

deanpienaar:
Related to Issue #793 

## Summary
* InlineRadios wasn't making correct use of `.form-check` and `.form-check-inline` as per the [Bootstrap Docs](https://getbootstrap.com/docs/4.0/components/forms/#inline)
* Changed `inline_class` to `is_inline` to better distinguish the differences
* Removed error partial as radios don't really have an invalid state (as seen [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#Providing_a_bigger_hit_area_for_your_radio_buttons)). Even bootstrap doesn't cover this case.
* Added `required` attribute to radios where applicable (which is normally the default case)

For screenshots and additional info, see #793

### DIFF
--- a/crispy_forms/templates/bootstrap4/layout/radioselect.html
+++ b/crispy_forms/templates/bootstrap4/layout/radioselect.html
@@ -2,8 +2,6 @@
 {% load l10n %}
 
 <div class="{% if field_class %} {{ field_class }}{% endif %}"{% if flat_attrs %} {{ flat_attrs|safe }}{% endif %}>
-    {% include 'bootstrap4/layout/field_errors_block.html' %}
-
     {% for choice in field.field.choices %}
       <div class="form-check{% if is_inline %} form-check-inline{% endif %}">
         <input type="radio" class="form-check-input"{% if choice.0|stringformat:"s" == field.value|default_if_none:""|stringformat:"s" %} checked="checked"{% endif %} name="{{ field.html_name }}" id="id_{{ field.id_for_label }}_{{ forloop.counter }}" value="{{ choice.0|unlocalize }}" {{ field.field.widget.attrs|flatatt }}>
--- a/crispy_forms/templates/bootstrap4/layout/radioselect.html
+++ b/crispy_forms/templates/bootstrap4/layout/radioselect.html
@@ -5,7 +5,7 @@
     {% include 'bootstrap4/layout/field_errors_block.html' %}
 
     {% for choice in field.field.choices %}
-      <div class="form-check{% if inline_class %} form-check-inline{% endif %}">
+      <div class="form-check{% if is_inline %} form-check-inline{% endif %}">
         <input type="radio" class="form-check-input"{% if choice.0|stringformat:"s" == field.value|default_if_none:""|stringformat:"s" %} checked="checked"{% endif %} name="{{ field.html_name }}" id="id_{{ field.id_for_label }}_{{ forloop.counter }}" value="{{ choice.0|unlocalize }}" {{ field.field.widget.attrs|flatatt }}>
         <label for="id_{{ field.id_for_label }}_{{ forloop.counter }}" class="form-check-label">
             {{ choice.1|unlocalize }}
--- a/crispy_forms/templates/bootstrap4/layout/radioselect.html
+++ b/crispy_forms/templates/bootstrap4/layout/radioselect.html
@@ -1,16 +1,16 @@
 {% load crispy_forms_filters %}
 {% load l10n %}
 
-<div class="{% if inline_class %}form-check{% endif %}{% if field_class %} {{ field_class }}{% endif %}"{% if flat_attrs %} {{ flat_attrs|safe }}{% endif %}>
+<div class="{% if field_class %} {{ field_class }}{% endif %}"{% if flat_attrs %} {{ flat_attrs|safe }}{% endif %}>
     {% include 'bootstrap4/layout/field_errors_block.html' %}
 
     {% for choice in field.field.choices %}
-      {% if not inline_class %}<div class="form-check">{% endif %}
-        <label for="id_{{ field.id_for_label }}_{{ forloop.counter }}" class="form-check-{% if inline_class %}{{ inline_class }}{% else %}label{% endif %}">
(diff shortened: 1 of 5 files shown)


crispy_forms/templates/bootstrap4/layout/radioselect.html
```diff
@@ -1,16 +1,14 @@
...
-
+<div class="{% if field_class %} {{ field_class }}{% endif %}"{% if flat_attrs %} {{ flat_attrs|safe }}{% endif %}>
     {% for choice in field.field.choices %}
-      {% if not inline_class %}<div class="form-check">{% endif %}
-        <label for="id_{{ field.id_for_label }}_{{ forloop.counter }}" class="form-check-{% if inline_class %}{{ inline_class }}{% else %}label{% endif %}">
-            <input type="radio" class="form-check-input"{% if choice.0|stringformat:"s" == field.value|default_if_none:""|stringformat:"s" %} checked="checked"{% endif %} name="{{ field.html_name }}" id="id_{{ field.id_for_label }}_{{ forloop.counter }}" value="{{ choice.0|unlocalize }}" {{ field.field.widget.attrs|flatatt }}>
+      <div class="form-check{% if is_inline %} form-check-inline{% endif %}">
+        <input type="radio" class="form-check-input"{% if choice.0|stringformat:"s" == field.value|default_if_none:""|stringformat:"s" %} checked="checked"{% endif %}{% if field.field.required %} required{% endif %} name="{{ field.html_name }}" id="id_{{ field.id_for_label }}_{{ forloop.counter }}" value="{{ choice.0|unlocalize }}" {{ field.field.widget.attrs|flatatt }}>
```
deanpienaar: Something I'm unsure of and might need guidance on is my addition of

```django
{% if field.field.required %}required{% endif %} 
```
on this line. I am not sure if it is the normal way it is done or if it is in the correct place. I was unable to find another case to base my addition off of.

deanpienaar:
Woops. Closing until I add the missing tests.

crispy_forms/templates/bootstrap4/layout/radioselect.html
```diff
@@ -1,16 +1,14 @@
...
-
+<div class="{% if field_class %} {{ field_class }}{% endif %}"{% if flat_attrs %} {{ flat_attrs|safe }}{% endif %}>
     {% for choice in field.field.choices %}
-      {% if not inline_class %}<div class="form-check">{% endif %}
-        <label for="id_{{ field.id_for_label }}_{{ forloop.counter }}" class="form-check-{% if inline_class %}{{ inline_class }}{% else %}label{% endif %}">
-            <input type="radio" class="form-check-input"{% if choice.0|stringformat:"s" == field.value|default_if_none:""|stringformat:"s" %} checked="checked"{% endif %} name="{{ field.html_name }}" id="id_{{ field.id_for_label }}_{{ forloop.counter }}" value="{{ choice.0|unlocalize }}" {{ field.field.widget.attrs|flatatt }}>
+      <div class="form-check{% if inline_class %} form-check-inline{% endif %}">
+        <input type="radio" class="form-check-input"{% if choice.0|stringformat:"s" == field.value|default_if_none:""|stringformat:"s" %} checked="checked"{% endif %}{% if field.field.required %} required{% endif %} name="{{ field.html_name }}" id="id_{{ field.id_for_label }}_{{ forloop.counter }}" value="{{ choice.0|unlocalize }}" {{ field.field.widget.attrs|flatatt }}>
```
deanpienaar: Something I'm unsure of and might need guidance on is my addition of

```django
{% if field.field.required %}required{% endif %} 
```
on this line. I am not sure if it is the normal way it is done or if it is in the correct place. I was unable to find another case to base my addition off of.

deanpienaar: @zoidbergwill

carltongibson: Hi @deanpienaar. This usage is fine. `BoundField` doesn't offer a `required` property, so what are you gonna do? 🙂

deanpienaar:
Sorry for the delay @carltongibson. Sure, I'll get on that.

deanpienaar:
Alright, sorted the checkboxes too


Feel free to let me know if there are any changes you'd like, @carltongibson

pbanaszkiewicz:
Hello, @deanpienaar and @carltongibson:

> * Removed error partial as radios don't really have an invalid state (as seen [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#Validation)). Even bootstrap doesn't cover this case.

I actually would like to ask you to bring it back. While Mozilla makes a good point, it's worth to notice that one can have two completely different fields, e.g. a numeric input and radio select, and custom form validation that ties them together, but `ValidationError` is only triggered on the radio. Additionally, [Bootstrap has validation styles for both radio selects and checkboxes](https://getbootstrap.com/docs/4.1/components/forms/#supported-elements).

In my case, I have a multiselect and a checkbox. User mustn't tick-off the checkbox unless a specific option is selected in the multiselect. Custom form validation raises `ValidationError` on the checkbox (`clean_checkbox` method).

Ref. #826.

deanpienaar:
Apologies. I've not found the time to implement this yet and this PR is probably falling behind. I'll close, and reopen a fresh one once I manage to get to this!

akramhussein:
Hi @deanpienaar - is this PR dead or has it been re-opened elsewhere, can't find it. Happy to help resolve it.

carltongibson:
Hi @akramhussein. If you want to pick this up, that would be amazing. 

More generally, if you're using crispy forms with BS4, we have a load of template issues which aren't hard per se, but need someone who knows (i.e. is using) BS4 to verify them. I just haven't got the bandwidth to do that. If you (or anyone reading this) could step up to fill that role I'd be very happy to mentor on the actual changes! And, if that happened we could get the BS4 support over the final hurdle to where it should be... 🙂

akramhussein:
@carltongibson I don't mind but I don't know much about BS4 and also don't have much bandwidth. Will take a look and let you know. Where would I start to find the issues?

Issue #734, opened 2017-08-22
Bootstrap 4: Prettier File Inputs


zoidbergwill:
File inputs don't look too great

![image](https://user-images.githubusercontent.com/2572493/29567161-a03a5676-874c-11e7-9245-f859f28995e6.png)

They should look more like:

![image](https://user-images.githubusercontent.com/2572493/29567193-b3ad9d9e-874c-11e7-9368-477ec6672a70.png)

Ours:

```
<input type="file" name="photo" class="clearablefileinput" id="id_photo">
```

What it should be closer to:

```
<label class="custom-file">
  <input type="file" name="photo" id="id_photo" class="custom-file-input">
  <span class="custom-file-control"></span>
</label>
```

csoni111:
I am new to this project and would like to contribute. As far as I have understood, `django-crispy-forms` does not handle the rendering of the field widget, `django` handles that. Now by default django renders the file widget as given above. So how do go on changing that?

zoidbergwill:
So we provide the template that Django uses to render the field widget in [`field.html`](https://sourcegraph.com/github.com/django-crispy-forms/django-crispy-forms@86b2991e2d6fba84e106ea792076aedee97343a4/-/blob/crispy_forms/templates/bootstrap/field.html) for each template pack, and then do special things for things like [radio box selects](https://github.com/django-crispy-forms/django-crispy-forms/blob/86b2991e2d6fba84e106ea792076aedee97343a4/crispy_forms/templates/bootstrap/layout/radioselect.html).

We could do the same for file inputs using our [`is_file` template tag](https://github.com/django-crispy-forms/django-crispy-forms/blob/a4afab25159f229ebc667834609d5ab52d31a491/crispy_forms/templatetags/crispy_forms_field.py#L41). Checking if it's a file, and then using that custom HTML above for Bootstrap 4 instead of the `crispy_field` template tag, potentially in a `layout/fileinput.html` template and not the main `field.html` template.

[...]

PR #809, opened 2018-04-03
Custom file input field for Bootstrap 4

fabianallendorf:
### Why merge this PR?
The Issue #734 suggested to add custom file input for Bootstrap 4.

### Description
A new layout template called `file_input.html` was created which holds the HTML of the Bootstrap4 [custom file input](https://getbootstrap.com/docs/4.0/components/input-group/#custom-file-input)
A distinction was included in `field.html` which checks if the field to render is a file input field using `is_file`.
The classes `form-control` and `is-invalid` are added to the field input in order to show form errors for the field.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 .PHONY: develop test
 
 develop:
-	python -m pip install -q -r requirements.txt
-	python -m pip install -q -e .
+	pip install -q -r requirements.txt
+	pip install -q -e .
 
 test: develop
 	DJANGO_SETTINGS_MODULE=crispy_forms.tests.test_settings py.test crispy_forms/tests --cov=crispy_forms
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 .PHONY: develop test
 
 develop:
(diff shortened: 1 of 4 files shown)


zoidbergwill:
Yay, a first-time contributor. 🎉 

Congratulations and thank you for your contribution, this looks great. Just gonna give it a test and then will merge it, since it looks good to me! :+1:

zoidbergwill:
Works perfectly, and looks much better. ✨

kantmichel:
Any time this will be implemented? Right now the file upload is blanking/not showing at all.when using standard implementation.

rafiqhilali:
I've been playing around with this, it may be worth adding to the docs that unless you add some custom JS, the label that says 'Choose File' will not update on choosing a file, leading to a slightly odd UI.

I added (using JQuery)
```javascript
const fileInput = $('#id_of_file_input');
  fileInput.change(function() {
    let fileName = fileInput.val();
    fileName = fileName.replace(/^.*[\\\/]/, '');
    fileInput.next('.custom-file-label').html(fileName);
  });

PR #811, opened 2018-04-06
I think we want this all in master

zoidbergwill:


zoidbergwill:
Awesome, I thought something like that might be happening. Sounds great. :+1: 👌

PR #820, opened 2018-06-01
Fixes #819: bootstrap 4 (and 3) select field with appended or prepended text not working properly

### LINKED ISSUE #819: bootstrap 4 Appended Text not working as expected for select fields
* Package version: 1.7.2
* Django version: 1.11.13
* Python version: 3.6.5
* Template pack: bootstrap4

### Description:

When appending a bootstrap.AppendedText to a layout for a select field, I expect to get something like the second example shown here:

https://getbootstrap.com/docs/4.0/components/input-group/#custom-select

This shows bootstrap 4 supports appending text to select inputs.

Instead the html generated by crispy-forms is missing both the div .input-group-append and the enclosing div .input-group.

The bug seems to be in `crispy_forms/templates/bootstrap4/layout/prepended_appended_text.html` which I intend to fix.

###  Example Crispy Forms code to reproduce

```
def __init__(self):
  self.helper = FormHelper()

  self.fields['select'] = forms.ChoiceField(label="Sample Label", help_text="Some help text", choices=[(1, "Choice 1"), (2, "Choice 2")])  # Choice fields are rendered with a select widget

  self.helper.layout = Layout(
    AppendedText('select', 'Some appended text')
  )
```

ociule:
I'll create a pull request for crispy-test-project that allows easily testing this patch.

### DIFF
--- a/crispy_forms/templates/bootstrap3/layout/prepended_appended_text.html
+++ b/crispy_forms/templates/bootstrap3/layout/prepended_appended_text.html
@@ -12,17 +12,11 @@
         {% endif %}
 
         <div class="controls {{ field_class }}">
-            {% if field|is_select %}
-                {% if crispy_prepended_text %}<span class="input-group{% if active %} active{% endif %}{% if input_size %} {{ input_size }}{% endif %}">{{ crispy_prepended_text|safe }}</span>{% endif %}
+            <div class="input-group">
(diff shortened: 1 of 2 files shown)


zoidbergwill:
This looks good to me. Thanks for your contributions! Sorry for the delay getting them merged.

PR #836, opened 2018-10-03
Allow field_template to be changed with template packs

### LINKED ISSUE #834: FormHelper.field_template does not work with template packs
* Package version: 1.7.2
* Django version: 2.0.4
* Python version: 3.5.2
* Template pack: 'bootstrap4' and 'bootstrap'

### Description:

I want to use the bootstrap template pack but overwrite the `field.html` template for customization (i.e. I want to hide the required asterisk only for select fields). However, setting the FormHelper's field_template attribute to "my_app/field.html" does not work. Looking into the the bootstrap templates I found that they do not consider field_template at all. This might be on purpose but it's not clear from reading the documentation.

So I'd suggest to either clarify in the docs what `field_template` does exactly or even better to make it behave as I expected: it overwrites where the template looks for the `field.html` template.

I managed to make it work by putting my custom `field.html` into `project_dir/my_app/templates/<template_pack_name>/field.html`.

robsdedude:
Fixes #834

### DIFF
--- a/crispy_forms/templates/bootstrap/uni_form.html
+++ b/crispy_forms/templates/bootstrap/uni_form.html
@@ -6,6 +6,6 @@
         {% include "bootstrap/errors.html" %}
     {% endif %}
     {% for field in form %}
-        {% include "bootstrap/field.html" %}
+        {% include field_template|default:"bootstrap/field.html" %}
     {% endfor %}
 {% endspecialspaceless %}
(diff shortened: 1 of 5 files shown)


robsdedude:
I was a little unsure whether to make a change in the templates or in the code. The changed branch in the templates only gets executed if no form layout is given explicitly. Another solution would have been to change the python code to enforce a layout (e.g. by creating a default layout in the render method).

zoidbergwill:
Thanks for your contribution, @robsdedude.

Looks good to me. :+1:

PR #870, opened 2019-04-10
Fixes Bootstrap4 issues #869 related to **Multiple** checkboxes. 

bryan-brancotte:
Fixes https://github.com/django-crispy-forms/django-crispy-forms/issues/869

rearranging the template so:
- each form-check-input is tagged as is-invalid
- input are not rendered in the label but before
- error is displayed after the label of the last element


bryan-brancotte:
When using a ChoiceField with a RadioSelect as widget the error is was not rendered, fixing it by applying same methodo as https://github.com/django-crispy-forms/django-crispy-forms/pull/870/commits/8b947deec95e801ceb6f8f1f41cf963d54ee7bbb : 
- each form-check-input is tagged as is-invalid
- input are not rendered in the label but before
- error is displayed after the label of the last element

PR #877, opened 2019-04-21
Fix tests for current pytest version

voidus:


johnthagen:
@voidus Could you also test against Django 2.2? Then I can just close #867 in favor of this.

I'd also suggest updating the `setup.py` trove classifier for 3.7 per my PR now that 3.7 is being tested.

voidus:
@johnthagen I added django 2.2 and python 3.8. I also took the liberty to remove python 3.4, which is unsupported since march.

PR #878, opened 2019-04-22
Bs4 offset checkboxes

voidus:
This includes #877 for working tests. If #877 is merged before this one, the diff will be nicer, but I can also isolate the relevant changes if that is preferred.

### DIFF
--- a/crispy_forms/helper.py
+++ b/crispy_forms/helper.py
@@ -372,10 +372,11 @@ class FormHelper(DynamicLayoutHandler):
         }
 
         if template_pack == 'bootstrap4':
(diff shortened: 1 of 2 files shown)


smithdc1:
Hi, I've rebased this on top of master so we can see the changes remaining changes.  Let's see if we can review this and get it merged? 👍

smithdc1:
They say an image can say a thousand words. So here we are:

Current release (1.8.0) with no additional column styling
180 no col

Current release (1.8.0) with additional column styling
180 with col

This PR with additional column styling
PR with col

This PR without additional column styling
PR no col

PR #880, opened 2019-05-06
Add ability to generate other tags than "div" to the "Div" layout class.

fsbraun:
The layout classes are a very versatile way of formatting forms. The ``Div`` layout class can be used to group elements and give them specific layouts (through the css_class parameter). Sometimes, however, it is helpful to have other elements, too, e.g. titles (``h1`` ... ``h6``). 

This pull requests adds a parameter tag to the ``Div`` class allowing the user to set the specific html tag required. If omitted it obviously falls back to the div tag.

This allows to keep HTML tags out of the ``HTML`` class making forms easier to internationalize

    Div(HTML(_('My great form')), tag='h1')

The required adaptations are minimal and I have done them for all template sets.

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -337,7 +337,8 @@ class MultiField(LayoutObject):
 
 class Div(LayoutObject):
     """
-    Layout object. It wraps fields in a <div>
+    Layout object. It wraps fields in a <div>, or alternatively any other tag you
+    specify with tag='h1'.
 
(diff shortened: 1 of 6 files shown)


zoidbergwill:
I can understand why this might be useful, especially in the future with custom web components

Can you add some tests for this feature?

My only other concern is maybe we should change the name of this class from Div to Element or something, since it's no longer always a div? We can alias from Div to Element and keep `div` as the default tag potentially to make migrations easier.

PR #884, opened 2019-06-08
Fix margin on error list

jacoduplessis:
Before:

wrong

After:

correct


bryan-brancotte:
There was the same problem in errors_formset.html, I applied you solution.

PR #892, opened 2019-08-23
Fix IDs of radio selects

FreeSalad:
https://github.com/django-crispy-forms/django-crispy-forms/pull/758 introduced wrong ID calculations for checkboxes and radio selects, resulting in `id_id_htmlname_0_index` instead of the expected `id_htmlname_index` of stock Django.

https://github.com/django-crispy-forms/django-crispy-forms/pull/783 has fixed this for checkboxselectmultiple already.

This PR fixes it for radio selects.

### DIFF
--- a/crispy_forms/templates/bootstrap/layout/radioselect.html
+++ b/crispy_forms/templates/bootstrap/layout/radioselect.html
@@ -5,8 +5,8 @@
     {% include 'bootstrap/layout/field_errors_block.html' %}
 
     {% for choice in field.field.choices %}
-        <label for="id_{{ field.id_for_label }}_{{ forloop.counter }}" class="radio{% if inline_class %} {{ inline_class }}{% endif %}">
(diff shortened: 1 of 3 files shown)


zoidbergwill:
The tests seem to be failing for an unknown reason, but could you also add some tests for your changes?

FreeSalad:
I don't think I have time to write some tests right now, I fixed this for one of my projects and I'm really busy with it for now.

I'm not sure tests are mandatory for those though, it's not a behavior that is bound to change over time I reckon. One pitfall though is that it isn't possible afaik to perfectly replicate Django's ID generation algorithm in templates, most of it being implemented in renderer functions. These changes however make it work with the majority of configurations that don't use auto_id or change widgets in a significant way.

smithdc1:
I have made some progress with this but I am unsure of the appropriate etiquette on taking this forward (do I just open a new PR?). I have simply taken @FreeSalad 's code updated it for #893 (failing tests) and I have added some tests based on what was done for #783 . 

As always, appreciate your guidance!


**Reference**
[My Branch / Code](https://github.com/smithdc1/django-crispy-forms/tree/radio_id-test)

Old HTML
Capture4

New HTML
Capture5

PR #915, opened 2019-10-21
Reworking layout.Row so it uses only template

bryan-brancotte:


smithdc1:
@bryan-brancotte looks good to me 👌 , shall I squash and merge?

bryan-brancotte:
yes if remarques in #916 do not apply here

PR #916, opened 2019-10-22
Reworking layout.Column so it uses only template + bugfix

bryan-brancotte:
All the css class are out of layout.py in independant html template
=> Add support of column out-of-the-box for BS4
Bugfix
=> remove the usage of css class `formColumn` in BS3 & BS4

Before
Capture d’écran de 2019-10-22 13-53-37

After
Capture d’écran de 2019-10-22 13-54-05

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -387,11 +387,16 @@ class Row(Div):
 
 class Column(Div):
     """
-    Layout object. It wraps fields in a div whose default class is "formColumn". Example::
+    Layout object. It wraps fields in a div so the wrapper can be used as a column. Example::
 
         Column('form_field_1', 'form_field_2')
+
+    Depending on the template, css class associated to the div is formColumn, row, or nothing. For this last case, you
+     must provide css classes. Example::
+
+        Column('form_field_1', 'form_field_2', css_class='col-xs-6',)
     """
-    css_class = 'formColumn'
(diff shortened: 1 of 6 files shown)


smithdc1:
Hi @bryan-brancotte a couple of questions: 

- This change and the change to row will presumably break other 3rd party templates that use this. However, as you found a bug when writing this change it doesn't appear to be a well used feature! (so let's not worry)

- On your after screenshot the buttons seem out of alignment to the form. It looks like their is extra padding on the form. 

A few notes from bootstrap docs. 

1. Notes on padding [Source](https://getbootstrap.com/docs/4.0/layout/grid/#how-it-works)
> Rows are wrappers for columns. Each column has horizontal padding (called a gutter) for controlling the space between them. This padding is then counteracted on the rows with negative margins. This way, all the content in your columns is visually aligned down the left side.

2. A column div should be inside a row div. [Example](https://getbootstrap.com/docs/4.0/layout/grid/#equal-width)

bryan-brancotte:
Hi @smithdc1 

Even if add css class `formColumn` in bootstrap in a bug, indeed maybe it is used bu some people. We should keep it for BS3 for backward compatibilty. As our support of BS4 is recent (not in beta I mean), maybe we can get ride of it in BS4.

Yes the buttons are missaligned, a problem in formactions.html that use a row as container, but no col for the children, a PR will come soon

PR #920, opened 2019-10-23
Usability of FileField in bootstrap4

bryan-brancotte:
Use custom template for FileInput along the vanilla JS making it fully functional.

Use django template for widget ClearableFileInput which is used when the form is a ModelForm (see https://github.com/django-crispy-forms/django-crispy-forms/issues/918#issuecomment-545382695)


carltongibson:
But see my comments on #918

carltongibson:
I didn't really mean to close this... Sorry. Hit the wrong button. 

@bryan-brancotte if you want to re-open it after reading https://github.com/django-crispy-forms/django-crispy-forms/issues/918#issuecomment-545579556, please do. 🙂

carltongibson:
Hiya, I think it would fall down to the the `else` clause below no, generating the older _vanilla_ file input?

I can't quite decide/workout whether we should add this here? (Will re-open for now while we decide. As I say, didn't mean to close it anyway. 🙄)

smithdc1:
Sorry, think I understand now. :smile: I missed that (very?) `vanilla` is an option.

bryan-brancotte:
Hi all,

So field_file.html add the label needed by boostrap, and render the input following what django would do. But when we are not in a form, but a model form, a clearable_file_input is used as widget with this following template : https://github.com/django/django/blob/master/django/forms/templates/django/forms/widgets/clearable_file_input.html leading to UI problem as in https://github.com/django-crispy-forms/django-crispy-forms/issues/918#issuecomment-545368286

I tried to copy paste its content in field_file.html, prefixing all replacing all {{ widget..something}} with {{ field.field.widget..something}} but it is not working. Note that {{ field.field.widget }} is not null.
 For example {{ field.field.widget.type }} is empty, so I am guessing that I am not using the objects correctly.

Maybe @carltongibson you have hint on how to rewrite a template from scratch ?

smithdc1:
Yes. I was just about to say something similar. This includes the required js for non clearable file fields which is a better solution to what I had in my head. 

Couple of questions.
- it needs docs
- should we use the same handler attribute as radios/checkbox?

carltongibson:
I think one **helper** attribute is enough no? (We can take a PR to add more fine grained control...)

smithdc1:
P.s. I'm happy to write up something for the docs this evening if that would be helpful?

bryan-brancotte:
Hi @smithdc1 I tried a quick edition of the documentation, feel free to improve it !

I think that one attribut helper to control all custom UI is enough, after all these customisation are also together in the BS4 docs

carltongibson:
Right, sorry, sometimes it takes a while to come up with the right idea. 

We should be telling people that they can override the template if they need to: 

```
{% elif field|is_file and not field|is_clearable_file and use_custom_control  %}
     {% include 'bootstrap4/layout/field_file.html' %}
```

So the conditional I guess is right. 

Then, if they already have their own JS, they can just provide  'bootstrap4/layout/field_file.html' without it. Not a perfect solution, but good enough for the bug fix. (I think 🙂)

smithdc1:
You are both amazing. 😃 

I've had a go at updating the docs including adding a table of examples. 
Hope this helps. 

Capture

bryan-brancotte:
Wow it looks so nice ! :clap: :clap: @smithdc1 

@carltongibson @smithdc1 
I don't think the JS embeded in field_file.html will cause any problem when developpers have already added the JS as it just set the value from the hidden input to the rendered fake input, doing it two time is useless but not harmfull. We can thus forget about indicating that they should override the template.

smithdc1:
Super work Bryan. :pray:

bryan-brancotte:
Thanks !

PR #932, opened 2019-10-31
Uneditable trim invisible from select

### LINKED ISSUE #817: UneditableField renders all select options
* Package version: 1.7.2
* Django version: 1.11.13
* Python version: 3.4.5
* Template pack: bootstrap4

### Description:

UneditableField still renders entire select dropdown with all options.  

I have an UneditableField for an associated User to my model, and for both performance and security reasons, I don't want every user to be rendered to the page when all I want is the current value displayed.

smithdc1:
This fixes #817 but a PR was never raised. I'm raising now so we can discuss if we want to make this change. 
(fingers crossed on the tests)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,25 @@
-language: python
-
 sudo: false
 
+
+language: python
 python:
   - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
-env:
-  - DJANGO='django>=1.8.0,<1.9.0'
-  - DJANGO='django>=1.11.0,<2.0'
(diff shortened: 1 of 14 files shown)


smithdc1:
tests failing for python 2 as default args need to be before optional? 

def __init__(self, field, *args, trim_invisible_from_select=False, **kwargs)

smithdc1:
I've created a branch [here](https://github.com/smithdc1/django-crispy-forms/tree/uneditable_trim_invisible_from_select) which has rebased this on top of master and has made the amend to make it python 2 compatible. I'll raise a new PR with those amends.

PR #934, opened 2019-11-02
Uneditable trim invisible from select

### LINKED ISSUE #817: UneditableField renders all select options

smithdc1:
This work was done some time ago but a PR was never raised (that I can see). I've rebased the changes and made a tweak to maintain python 2 support. Fixes #817


smithdc1:
Hi @carltongibson. Thanks for your comments. 

Tests, yes. I'll take a look at that.

As for docs, this got me thinking. The solution that was proposed added a new helper, and so yes we should have docs for that. But I've now got some questions!

- it currently defaults to false, should this by true? I suspect that's what most people will expect.
- is there a use case to make want an uneditable field to be editable? Should we just implement this with out the helper option?

@bryan-brancotte appreciate your thoughts to.

bryan-brancotte:
Hi @smithdc1 

Maybe our approach to keep a select is not the good idea, as mentioned in https://github.com/django-crispy-forms/django-crispy-forms/issues/817#issuecomment-395691381 with boostrap2 it was only a span, so not all the option, maybe we should go back to this behavior and use https://getbootstrap.com/docs/4.0/components/forms/#readonly ? 

### Pros:
* Do not adding a new option in the helper
* Do not need documentation
* Only render the current value, not the others

### Cons:
* If a crispy user create field as read-only and then want to enable it in JS it is not possible

In conclusion : either user readonly of BS4 or just disabled plus not-triming

smithdc1:
Hi @bryan-brancotte 
I didn't quite understand your proposal. Are you suggesting that we either go back to bs2 style or that no change should be made? 

If so I suggest the latter, this is especially true if problem be solved by a method in a form class (similar to #935 ) ?


In the meantime I've had another look at this and writing some tests and docs helped me to understand this issue in more detail. Below is my thought process. 

I've added three fields to the test project being
- standard select
- disabled select (which generates the options in the HTML cost), 
- disable with trim option = `True`

See [here](https://wizardly-rosalind-ffa089.netlify.com/) for generated HTML. The fields are just above the submit buttons.

When setting the option to `True` the field disappears all together rather than outputting a `span`. I *think* this is because I need to pass the data to the form for it to match with options?? (sorry, run out of knowledge, again :-)). So question is, does this even work as it is? If it does can we write an appropriate test?  

Having written the docs, I don't think it is too intrusive. By default the behaviour is unchanged (no trimming) and so it allows people to enable the disabled field through JS. The trimming function is done through an argument on the `uneditable` field rather than another addition to the helper. This makes more sense with the longer term project of splitting out the template packs.

bryan-brancotte:
Hi @smithdc1 
I have mixed feeling about this issue, mostly because the initial need is for me wrong : asking crispy to add option to help not exposing the complet list of user is for me note the good approach, I think one should instead in the __init__ of the form trim the queryset for the field associated to User. When trimming in the __init__ the need of a trimming in the field never arise.

For the bs2 approach, if we want to let dev trim the queryset, I proposed https://getbootstrap.com/docs/4.0/components/forms/#readonly as it looks nice, but don't know the amount of work to make it work. Indeed the span is not printed so I guess it is, at least, not working out of the box. So maybe just forget my proposition and keep the trimmed select ;) ?

>By default the behaviour is unchanged (no trimming) 
:+1: 

> The trimming function is done through an argument on the uneditable field
:+1:

PR #945, opened 2019-11-15
Cleanup CI configuration

Djailla:
- Python 3.8 is now official
- Django 2.x is not officially compatible with Python 3.8
- Cleanup indent in tox.ini

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,10 @@
-sudo: false
-
-
-dist: xenial
 language: python
 python:
   - "2.7"
   - "3.5"
   - "3.6"
   - "3.7"
-  - "3.8-dev"
+  - "3.8"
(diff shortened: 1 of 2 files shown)


smithdc1:
Hi @Djailla 
Thanks for the PR 👍

The [Django docs](https://docs.djangoproject.com/en/2.2/releases/2.2/#python-compatibility) show Django 2.2 supports python 3.8 from version 2.2.8. 

Do you think you could make the change to test Django 2.2 against Python 3.8? We have a few other PRs open at the moment which change the same file, so would be good to rebase those on top of this!

Thank you!

Djailla:
@smithdc1 Done

smithdc1:
That's great, thank you 🙏

PR #949, opened 2019-11-22
Add GitHub actions for Flake8

smithdc1:
Here is the first small step towards exploring github actions. 

This PR enables flake8 only. To get the 'test' to pass I have had to amend a couple of test names (good spot flake8!?), and I've added some exceptions. 

I've had to take an opinion on the settings for flake8. I've copied GitHub's recommendations with the exception of line length. 

This leaves c.50 'errors' on the current code base. We will need to consider how to approach these.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -1,0 +1,34 @@
+name: Python package
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.7]
+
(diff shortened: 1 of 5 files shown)


smithdc1:
Hi @carltongibson 

I've started to rebase the PRs following removal of Python 2 support. 

In terms of priority, could you look at this PR next (I really don't mean to push, just trying to prioritise, so whenever you have time), as I this has the potential to 'unlock' a piece of work if we wish flake8 to pass with no 'errors'. There were c.50 errors when I ran it before the holiday season. 


As always appreciate your support / guidance.

PR #950, opened 2019-11-28
Improve render_hidden_fields docs fixes #886

### LINKED ISSUE #886: ModelForm with helper.layout does not render hidden pk field added by modelformset
* Package version: 1.7.2
* Django version: 2.2
* Python version: 3.7
* Template pack: Bootstrap

Using `{% crispy form %}` template tag in loop to render each form in a modelformset.
The underlying form class has a helper that defines a layout.
The form class does NOT list the model's pk field - modelformset_factory adds a hidden pk field to the form, as per usual.
But the hidden pk field is not rendered, causing the modelformset logic to bork on POST request.
If I comment out the helper.layout (i.e., use default layout), all works as it should.

Presumably this is because the layout logic has no way to know the hidden pk field needs to be rendered to make the modelformset save logic work correctly.
This seems like a documentation issue, and I'm happy to submit a PR to document this and save someone else the debugging effort, but hoping to confirm (i) this behaviour is as intended; and (ii) there is no way to render the hidden pk field that gets added to the modelformset forms?

### Preferably also include:

forms.py
```
class MyModelForm(forms.ModelForm):
    class Meta:
        model = models.MyModel
        fields = ( 'project_status',  )  # Note : id field NOT part of normal form

        super().__init__(*args, **kwargs)
        self.helper = FormHelper()
        self.helper.form_tag = False
        self.helper.form_show_labels = False
        self.helper.disable_csrf = True
        # The offending layout:
         self.helper.layout = Layout(
             Field('project_status', css_class='use-select2 no-search'),
         )
```

ModelFormset generated standard way:
```
ModelFormset = modelformset_factory(models.MyModel, **factory_kwargs)
...
formset = ModelFormset(queryset=qs, data=data, ...)
```

HTML generated using layout
```
<div id="div_id_form-0-project_status" class="form-group"> <div class="controls "> <select name="form-0-project_status" class="select form-control" id="id_form-0-project_status"> <option value="Completed" selected="">Completed</option> <option value="Archived">Archived</option>
```

HTML generated without layout:
```
<div id="div_id_form-0-project_status" class="form-group"> <div class="controls "> <select name="form-0-project_status" class="select form-control" id="id_form-0-project_status"> <option value="Completed" selected="">Completed</option> <option value="Archived">Archived</option>

</select> </div> </div> <input type="hidden" name="form-0-id" value="699" id="id_form-0-id">
```

smithdc1:
Issue #886 raised that the docs could be more clear on the purpose of the render_hidden_fields feature. I've added a few words to try and help in future. 

Closes #886

### DIFF
--- a/docs/form_helper.rst
+++ b/docs/form_helper.rst
@@ -90,7 +90,7 @@ Helper attributes you can set
     By default django-crispy-forms renders the layout specified if it exists strictly, which means it only renders what the layout mentions, unless your form has ``Meta.fields`` and ``Meta.exclude`` defined, in that case it uses them. If you want to render unmentioned fields (all form fields), for example if you are worried about forgetting mentioning them you have to set this property to ``True``. It defaults to ``False``.
 
 **render_hidden_fields = False**
-    By default django-crispy-forms renders the layout specified if it exists strictly. Sometimes you might be interested in rendering all form's hidden fields no matter if they are mentioned or not. It defaults to ``False``.
+    By default django-crispy-forms renders the layout specified if it exists strictly. Sometimes you might be interested in rendering all form's hidden fields no matter if they are mentioned or not. Useful when trying to render forms with layouts as part of a formset with hidden primary key fields. It defaults to ``False``.
 
 **render_required_fields = False**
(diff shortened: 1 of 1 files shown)


smithdc1:
This is a small docs change so merging as is.

PR #952, opened 2019-11-30
Render fields not included in the explicit layout in the order they are provided

### LINKED ISSUE #951: How do I specify the order of fields in a form?
* Package version: django-crispy-forms==1.7.2
* Django version: Django==2.1.14
* Python version: 3.6
* Template pack: CRISPY_TEMPLATE_PACK = 'bootstrap4'

### Description:

https://stackoverflow.com/questions/59101176/how-do-i-specify-the-order-of-fields-in-a-form-django-crispy-forms

We are using Django 2.1 for [Speedy Net](https://github.com/speedy-net/speedy-net). We have a contact form and recently it has been used by spammers to send us spam. I decided to add a "no_bots" field to the form where I'm trying to prevent bots from submitting the form successfully. I checked the form and it works, but the problem is we have 2 sites - on one site (Speedy Net) the order of the fields is correct, and on the other site (Speedy Match) the order of the fields is not correct - the "no_bots" field comes before the "message" field but I want it to be the last field. How do I make it last? Our template tag contains just `{% crispy form %}` and I defined the order of the fields in `class Meta`.

Our code:

```
class FeedbackForm(ModelFormWithDefaults):
	...
	no_bots = forms.CharField(label=_('Type the number "17"'), required=True)

	class Meta:
		model = Feedback
		fields = ('sender_name', 'sender_email', 'text', 'no_bots')

		super().__init__(*args, **kwargs)
		self.helper = FormHelperWithDefaults()
		if (self.defaults.get('sender')):
			del self.fields['sender_name']
			del self.fields['sender_email']
			del self.fields['no_bots']
			self.helper.add_input(Submit('submit', pgettext_lazy(context=self.defaults['sender'].get_gender(), message='Send')))
		else:
			self.fields['sender_name'].required = True
			self.fields['sender_email'].required = True
			self.helper.add_layout(Row(
				Div('sender_name', css_class='col-md-6'),
				Div('sender_email', css_class='col-md-6'),
			))
			self.helper.add_input(Submit('submit', _('Send')))

	def clean_text(self):
		text = self.cleaned_data.get('text')
		for not_allowed_string in self._not_allowed_strings:
			if (not_allowed_string in text):
				raise ValidationError(_("Please contact us by e-mail."))
		return text

	def clean_no_bots(self):
		no_bots = self.cleaned_data.get('no_bots')
		if (not (no_bots == "17")):
			raise ValidationError(_("Not 17."))
		return no_bots
```

Screenshots:

Speedy Net
Speedy Match

acjh:
Fixes #951 
Follows https://github.com/django-crispy-forms/django-crispy-forms/pull/539#issuecomment-307984618

The first loop is a superset of the second loop.
Let's combine the loops and use the order in Meta, followed by the order in form.

More info: [https://stackoverflow.com/questions/59101176/how-do-i-specify-the-order-of-fields-in-a-form-django-crispy-forms/59120457#59120457](https://stackoverflow.com/a/59120457/8601760)

### DIFF
--- a/crispy_forms/helper.py
+++ b/crispy_forms/helper.py
@@ -8,7 +8,7 @@ from crispy_forms.exceptions import FormHelpersException
 from crispy_forms.layout import Layout
 from crispy_forms.layout_slice import LayoutSlice
 from crispy_forms.utils import (
-    TEMPLATE_PACK, flatatt, list_difference, list_intersection, render_field,
+    TEMPLATE_PACK, flatatt, list_difference, render_field,
 )
 
 try:
@@ -313,11 +313,20 @@ class FormHelper(DynamicLayoutHandler):
         if self.render_unmentioned_fields or self.render_hidden_fields or self.render_required_fields:
             fields = tuple(form.fields.keys())
             left_fields_to_render = list_difference(fields, form.rendered_fields)
+
+            # If the user has Meta.fields defined, we use that order
+            if hasattr(form, 'Meta') and hasattr(form.Meta, 'fields'):
+                field_indexes = {field: index for index, field in enumerate(form.Meta.fields)}
+                end_index = len(form.Meta.fields)
+                left_fields_to_render = sorted(left_fields_to_render, key=lambda x: field_indexes.get(x, end_index))
+
             for field in left_fields_to_render:
+                # We still respect the configuration of the helper
+                # regarding which fields to render
                 if (
                     self.render_unmentioned_fields or
-                    self.render_hidden_fields and form.fields[field].widget.is_hidden or
-                    self.render_required_fields and form.fields[field].widget.is_required
+                    (self.render_hidden_fields and form.fields[field].widget.is_hidden) or
+                    (self.render_required_fields and form.fields[field].widget.is_required)
                 ):
                     html += render_field(
                         field,
@@ -326,28 +335,6 @@ class FormHelper(DynamicLayoutHandler):
                         context,
                         template_pack=template_pack
                     )
-
-        # If the user has Meta.fields defined, not included in the layout,
-        # we suppose they need to be rendered
-        if hasattr(form, 'Meta'):
-            if hasattr(form.Meta, 'fields'):
-                current_fields = tuple(getattr(form, 'fields', {}).keys())
-                meta_fields = getattr(form.Meta, 'fields')
-
-                fields_to_render = list_intersection(current_fields, meta_fields)
-                left_fields_to_render = list_difference(fields_to_render, form.rendered_fields)
-
-                for field in left_fields_to_render:
-                    # We still respect the configuration of the helper
-                    # regarding which fields to render
-                    if (
-                        self.render_unmentioned_fields or
-                        (self.render_hidden_fields and
-                         form.fields[field].widget.is_hidden) or
-                        (self.render_required_fields and
-                         form.fields[field].widget.is_required)
-                    ):
-                        html += render_field(field, form, self.form_style, context)
 
         return mark_safe(html)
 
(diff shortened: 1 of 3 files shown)


smithdc1:
Hi @acjh ,

Thanks for looking at this. 👍 

Do you think we should have a regression test?

acjh:
Hi @smithdc1,

Thanks for responding.

Yes, let me work on a regression test.

acjh:
@smithdc1 I've added two regression tests:

1. `test_render_unmentioned_fields` — unaffected by the changes in this MR
2. `test_render_unmentioned_fields_order`

While adding `test_render_unmentioned_fields_order`, I realised that #397 (released Oct 2017 in version 1.7.0) effectively regressed https://github.com/django-crispy-forms/django-crispy-forms/commit/df76363319efc9da1d51bfa900c4cc3bc2960511 (released Feb 2012 in version 1.1.0), making that loop completely unnecessary. Thus, I've backtracked from combining the loop (https://github.com/django-crispy-forms/django-crispy-forms/pull/952/commits/2ab4e951b2ae922d3ae30b8ebefe45735d2c1a6d) in https://github.com/django-crispy-forms/django-crispy-forms/pull/952/commits/1fbb0c8920f36b928a90e04194fed0b39154f216.

smithdc1:
Hi @acjh 
Thanks again for your work on this. 

So are you saying that since 1.7.0 fields that have **only** been mentioned in Meta have not been rendered. e.g in the below example I don't see the model fields. 

If so, is this something we should fix. I don't think #397 intended to remove this feature?

```
class MessageForm(ModelForm):
    class Meta:
        model = Musician
        fields = ['first_name', 'last_name', 'instrument']

    radio_buttons = forms.ChoiceField(
        choices=(
            ('option_one',
             "Option one is this and that be sure to include why it's great"),
            ('option_two',
             "Option two can is something else and selecting it will deselect option one")
        ),
        widget=forms.RadioSelect,

    )

    select_field = forms.ChoiceField(
        choices=[(1, "Choice 1"), (2, "Choice 2"), (3, "Choice 3")],
    )

    helper = FormHelper()
    helper.layout = Layout(
        'radio_buttons',
        'select_field',
    )
```

If we set render_unmentioned_fields = True the fields of course are then mentioned, and with your PR they are in the correct order. 

Old 
Capture

New
Capture2


Finally, the test's look good to me

acjh:
> So are you saying that since 1.7.0 fields that have **only** been mentioned in Meta have not been rendered. e.g in the below example I don't see the model fields.

Yes, since 1.7.0 (released Oct 2017).

> If so, is this something we should fix. I don't think #397 intended to remove this feature?

I do agree that "fields that have **only** been mentioned in Meta" should be rendered. Yet #397 did as stated in its description and that has been the behaviour for over 2 years. I feel that @carltongibson, who merged #397 and is still a maintainer, should weigh in on whether it is something intended now.

smithdc1:
Super, I think you've got this to a point where we're now making a judgment call on how this should work. 

I've also been reflecting on this and my thoughts are:

- I can't see an issue raised on it (rendering fields only mentioned in Meta, even when render_unmentioned_fields is not set; default = False) not working. Given this was included in a release a couple years ago it doesn't look like a well used feature. 

- There was an old issue to do the opposite (i.e. not render meta fields). https://github.com/django-crispy-forms/django-crispy-forms/issues/316

Given that explicit is better than implicit and no one has shouted I suggest we merge this PR as is. 

Happy to hear @carltongibson or @bryan-brancotte before doing so. 😄

acjh:
> - There was an old issue to do the opposite (i.e. not render meta fields). #316

Ah, nice find.

smithdc1:
Thanks both - lets merge 👍

PR #955, opened 2019-12-05
Update layouts.rst

soulshake:
This change clarifies the import path for `UneditableField`, which I had a hard time figuring out how to import, since it's in `crispy_forms.bootstrap` (instead of `crispy_forms.layout` like `Field`, `Fieldset`, etc. are).


soulshake:
Never mind, I see now there's a note [here](https://django-crispy-forms.readthedocs.io/en/latest/layouts.html#bootstrap-layout-objects) describing the import path :woman_facepalming:

Issue #960, opened 2020-01-05
Translation does not work for button used within common layout


monim67:
* Package version: 1.8.1
* Django version: 3.0.1
* Python version: 3.7.6
* Template pack: (Optional)

### Description:

Translation does not work for button text when used within a common layout as described in the docs [here](https://django-crispy-forms.readthedocs.io/en/latest/layouts.html#composing-layouts). The translation does not update as it should when the language is changed.

### Steps to reproduce:

- Go to [this heroku app](https://django-crispy-forms-lazy-issue.herokuapp.com/en/).
- Switch the language using language selection links at the bottom.
- Language updates for the whole page except for the button text.

See the source code of it [here on GitHub](https://github.com/monim67/django-crispy-forms-lazy-issue/blob/f544a4c/btn_translation_issue/forms.py#L7-L11) using common layout for buttons.

### Why this happens:

This happens because when `BaseInput` class of crispy-forms renders it converts the lazy django translation object passed as `value` to string and stores it back to `value` overwriting it [here](https://github.com/django-crispy-forms/django-crispy-forms/blob/b55b86a/crispy_forms/layout.py#L197), hence the lazy translation object is lost upon the first render of the input element. If we don't overwrite the original `value` the issue is resolved.

tobiasmboelz:
This is not limited to the buttons in a common layout. The problem exists for all `Button`s/`Submit`s used in any layout.

PR #961, opened 2020-01-05
Added value_template_string property to BaseInput

monim67:
### Fixes:
- Button translation issue #960.

### What has been changed:
- Added a new property `value_template_string` to `BaseInput` class.

### How it fixes the issue:
The additional property `value_template_string` preserves the original value passed to sub-classes
of `BaseInput` from being overwritten in the `render` method.


smithdc1:
Hi @monim67 - thanks for looking into this! 

Do you think we should have a small regression test?

monim67:
@smithdc1 how can we do that? So far the tests run successfully so and I have checked the input renders properly after this change. Do you have anything on mind?

smithdc1:
I had a look at your heroku demo app which I think you've updated for the change as the buttons now change language.

In this version it doesn't look like it's working as intended. Whilst the labels change language ok, the button behaviour is a bit random. 

ezgif-1-9b562fe18507

monim67:
The heroku app is intended to demonstrate the bug, it does not include the fix. And the button labels don't change language. You might be seeing a cached version there.

carltongibson:
Traditionally we've said that if you need translation then you **must** instantiate your layout objects in a per-request context, such as a form's `__init__`. You can wrap the shared layout in a helper/factory function to keep it DRY. 

Making the values properly translatable would be a reasonable addition I think. But it would be a new feature, and I think we should make sure we've thought it through fully before just sneaking it in. (Including tests, and documenting any new attributes. 

c.f. #776 here, which looked at the same issue. (Not sure if the approach was sound or not as yet.) 

An alternate approach I was pondering whilst thinking about this, but haven't thought through, is finding a way of not needing to overwrite value each time through render... (From the second pass it's a noop — can we lazily evaluate it somehow? — Not sure as yet.)

monim67:
@carltongibson while looking into it the lazy object also came to my mind, but if I assign a lazy object to `value` in the constructor then it will require the `context` object from render function to evaluate, so it's not going to work unless we find a way to pass the `context`.

The simple solution I thought was separating two value variables instead of overwriting, the one passed to the constructor `value_template` and the one passed to html template `rendered_value`. I was in a confusion about which one to rename. I chose to rename the `value_template` considering the affect was less.

We should really think deeper before we decide how to provide the feature. Haul me if I can be of any help 👍

PR #964, opened 2020-01-14
Ref #926 Fix class on File upload input field.

CormacMOB:
Bootstrap 4 uses the "form-control-file" class for file input fields but the "form-control" class was incorrectly being applied to file input fields leaving a field border that is not present in Bootstrap 4 Documentation.
This will append to the class to correct the class string.

### DIFF
--- a/crispy_forms/templatetags/crispy_forms_field.py
+++ b/crispy_forms/templatetags/crispy_forms_field.py
@@ -135,6 +135,8 @@ class CrispyFieldNode(template.Node):
             ):
                 if not is_checkbox(field):
                     css_class += ' form-control'
+                    if is_file(field):
+                        css_class += '-file'
                 if field.errors:
                     css_class += ' is-invalid'
 


smithdc1:
Hi @CormacMOB,

Thanks for looking at this. 

We now have form-control-file on the file `<input>` tags. This means there is no border on the 'vanilla' file inputs. 

For the custom file inputs we now have the rather long css class of

`class="fileinput fileUpload form-control-file custom-file-input"`

Do you think it's ok to have both form-control-file and custom-file-input? It looks ok...
screencapture-127-0-0-1-8000-2020-01-14-22_46_26

CormacMOB:
No worries @smithdc1 

The form-control-file class doesn't do a lot, intentionally I think.

The "custom-file-input" class agrees mostly with "form-control-file" the only difference is that th former applies display: inline-block instead of display: block. Given that width is defined as 100% by the "custom-file-input" class I don't think this is a problem.

I looked at it on Chrome and FF and poked at it on both the BS4 doc site and the Crispy Test app and it looks as if it is safe to apply both classes on the custom file input.

smithdc1:
ok - thanks. 🙏

PR #970, opened 2020-01-18
Drop Django V1 support

smithdc1:
To remove compatibility for Django version 1.

The one bit which needed further research as it's not documented in the code is `_get_new_csrf_string`. Django docs show 'string' in [django version 2.2](https://docs.djangoproject.com/en/2.2/_modules/django/middleware/csrf/) and 'key' in [django version 1](https://docs.djangoproject.com/en/1.8/_modules/django/middleware/csrf/). I've therefore removed the option to try for 'key'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 * Removed support for Python 2
+* Removed support for Django version 1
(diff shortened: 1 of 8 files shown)


smithdc1:
Thanks Carlton 👍

PR #974, opened 2020-01-18
flake8 refactoring

smithdc1:
A few more small changes as highlighted by flake8.


smithdc1:
I think most of the easier changes have now been made. I'll start to think about a point release now. This work can be revisited later.

PR #976, opened 2020-01-24
Improve multivalue support

smithdc1:
This is some early work towards fixing #831 . This isn't much more than a proof of concept as there are many more things to consider, along with tests, documentation etc. 

However, I'd appreciate views on if this is a reasonable approach to this problem and if it is worth exploring further. 

A few notes:

1. field.html has been updated to use multifield.html

2. a new template tag has been created. Most of it is copied from [here](https://github.com/treyhunner/django-widget-tweaks/blob/multi-widgets/widget_tweaks/templatetags/widget_tweaks_extras.py) with a few modifications to get it working for my requirements. This tag allows us to loop over each field within a multifield.

3. multifield.html iterates over the multifield allowing us to put each `<input>` within a `<div>` allowing us to add a col class for bs4 style.

### DIFF
--- a/crispy_forms/templates/bootstrap4/field.html
+++ b/crispy_forms/templates/bootstrap4/field.html
@@ -25,7 +25,7 @@
         {% endif %}
                 
         {% if field|is_multivalue %}
-                {% include 'bootstrap4/layout/multifield.html' %}
+                {% include 'bootstrap4/layout/multivalue.html' %}
         {% endif %}        
 
         {% if not field|is_checkboxselectmultiple and not field|is_radioselect and not field|is_multivalue %}
--- a/crispy_forms/templates/bootstrap4/field.html
+++ b/crispy_forms/templates/bootstrap4/field.html
@@ -23,8 +23,12 @@
         {% if field|is_radioselect %}
             {% include 'bootstrap4/layout/radioselect.html' %}
         {% endif %}
+                
+        {% if field|is_multivalue %}
+                {% include 'bootstrap4/layout/multifield.html' %}
+        {% endif %}        
 
-        {% if not field|is_checkboxselectmultiple and not field|is_radioselect %}
+        {% if not field|is_checkboxselectmultiple and not field|is_radioselect and not field|is_multivalue %}
             {% if field|is_checkbox and form_show_labels %}
                 {%if use_custom_control%}
                     {% crispy_field field 'class' 'custom-control-input' %}
--- a/crispy_forms/templates/bootstrap4/layout/multifield.html
+++ b/crispy_forms/templates/bootstrap4/layout/multifield.html
@@ -1,5 +1,4 @@
 {% load crispy_forms_field %}
-{% load crispy_forms_tags %}
 
 {% if field.is_hidden %}
     {{ field }}
@@ -13,20 +12,12 @@
         {% crispy_field field %}
     {% endif %}
 
-
+    {% if field.label %}
+        {{ field.label }}
+    {% endif %}
 
     {% if not field|is_checkbox %}
-            <div class="row {% if field_class %} {{ field_class }}{% endif %}">
-            {% for f in field.field.widget.widgets %}
-            <div class="col-md">
-
-                {% render_multi_field field forloop.counter0 'class' 'form-control' %}
-           </div>
-                
-            {% endfor %}
-            </div>
-
-            
+        {% crispy_field field %}
     {% endif %}
 
     {% if field.label %}
--- a/crispy_forms/templates/bootstrap4/layout/multifield.html
+++ b/crispy_forms/templates/bootstrap4/layout/multifield.html
@@ -16,7 +16,7 @@
 
 
     {% if not field|is_checkbox %}
-            <div class="row">
+            <div class="row {% if field_class %} {{ field_class }}{% endif %}">
             {% for f in field.field.widget.widgets %}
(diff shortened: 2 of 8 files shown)


smithdc1:
I've had a few thoughts on this over the weekend. 

There is duplication in render_multi_field and crispy_field, particularly the way that the additional attributes are handled. They also do this in different ways, we should at least adopt the crispy_field way of doing things, I think. I feel consistency is most important. 

Do we add the additional 'looping' functionality to crispy_field - maybe through an additional attribute which we can catch and do something with. How do we do this in a way which is backwards compatible. 

Multifield.html needs to be able to render different code depending what field is within the multifield. Most may be straight forward `textinput` but if there is a radio or checkbox then additional `div`s are needed and the css on `<input>` is different also. 

This will be a fairly big change to the multifiled style, so should we need to add a flag for backwards compatibility.

smithdc1:
I've taken the approach to base `MultiFieldAttributeNode` on much of the `CrispyFieldNode` with the added functionality to render the individual field that is passed into the function. This means there is a fair amount of duplication, but style of passing in attributes is consistent and should be easier to merge the two functions if we decide that is ok.

I've also stripped back the additional CSS styles in this function, we should pass it in via the template. This is a slightly wider point which I'll come back to post 1.9.0. I'd like to make some more progress on using the template packs to pass in CSS styles rather than have them coded into the core application. This will make it easier to decouple the template packs.

smithdc1:
As per #591 there are two multifield templates, one of which is missing and worse I've interpreted the existing one as the 'field' template rather than the 'layout object' template. Therefore we need to:

- [ ] move current multifield.html file up a folder level
- [ ] bring in an equivalent of the [bs3 template](https://github.com/django-crispy-forms/django-crispy-forms/blob/master/crispy_forms/templates/bootstrap3/layout/multifield.html)

I've done some more work on my old device. I'll make that commit and then come back with a to-do list to complete the 'field' template.

smithdc1:
I've made some more progress on this, and I've learnt that 'mulivalue' and 'multifield' has caused me quite some confusion. I'll delete the previous comment as that turned out to be completely wrong!

What I've been trying to solve here is 'multivalue' fields. e.g. [SplitDateTimeField](https://docs.djangoproject.com/en/3.0/ref/forms/widgets/#splitdatetimewidget) and this is completely different to the multfield layout object in crispy-forms. 

I've therefore created a new template for 'multivalue' fields. In addition to what I see as the standard usage of multivalue fields I've also added some support for a single checkbox. However, I've not gone as far as allowing multiple checkbox / radio within a multivalue field. 

Whilst I've got a couple of remaining tests to fix, here are some images of what it currently looks like. 




Capture3
Capture2
Capture

smithdc1:
Hi @carltongibson please could I have some guidance. 

I'm struggling to understand what is happening with this test. 

https://github.com/django-crispy-forms/django-crispy-forms/blob/77f4ea7edd3d8a4cd21f20bbe7fbce1511248e10/crispy_forms/tests/test_layout_objects.py#L55

The html being produced currently is as per below (bs3 template pack). There are two input fields being produced for the `splithiddendatetimewidget` and therefore three inputs in total for a `SplitDateTimeField`. (i.e. a date field, a hidden date field and a hidden time field)

Do you have a view on what the expected behaviour is here? 

```
<form method="post">
    <div id="div_id_datetime_field" class="form-group">
        <label for="id_datetime_field_0" class="control-label  requiredField">
            date time<span class="asteriskField">*</span> </label>
        <div class="controls ">
            <input type="text" name="datetime_field_0" rel="test_dateinput" class="dateinput" required id="id_datetime_field_0">
            <input type="hidden" name="datetime_field_1_0" rel="test_timeinput" style="width: 30px;" class="splithiddendatetimewidget" required id="id_datetime_field_1_0">
            <input type="hidden" name="datetime_field_1_1" rel="test_timeinput" style="width: 30px;" class="splithiddendatetimewidget" required id="id_datetime_field_1_1"> </div>
    </div>
</form>
```

sbutler:
I was thinking very long and hard about this problem, and started to wonder why in `crispy_forms.templatetags.crispy_forms_field.CrispyFieldNode#render` you loop over the widgets, but the test for checkboxes is on the field. Wouldn't it solve all the problems if the checkbox test is just on the widget itself?

I did a [proof-of-concept code](https://github.com/sbutler/django-crispy-forms/tree/1.8.1-multiwidget) that does two things:

1. Modifies the various `is_XXX` tests to accept any of `BoundField`, `Field`, or `Widget`.
2. Changes the loop's tests so that it is on the current widget and not the field as a whole.

I'm still working on it, but for me the initial results here are what I'd want and I think it solves the issue of adding 'form-control' to widgets.

smithdc1:
Hi @sbutler Thank you for thinking about this. 👍 

I also have some thoughts on the piece of code you are referring too (where we add `form-control` and `is-invalid`). One of the aims we have as a project is to decouple the template packs and the core logic. This is one area where the two are tightly combined. 

I'd therefore prefer we removed this logic from `crispy_forms.templatetags.crispy_forms_field.CrispyFieldNode#render` and add it into the template packs. We can do this by passing css_class attributes into crispy_field from each template. 

Appreciate your thoughts.

smithdc1:
Note to self.

I wonder if we can loop to a sub template for each sub field rather than trying to fix this in python code.

PR #980, opened 2020-02-06
Only show MultiField if fields have errors

mulhollandms:
I noticed that when using the 'bootstrap3' template pack, the MultiField layout object shows the alert box for errors regardless of whether there actually are any. Based on the 'layouts/multifield.html' template, it seems the only prerequisite for showing the alert is that "form_show_errors" is enabled.

This was just a quick patch to solve my problem without disabling form errors entirely so I'm sure it'll be worth another look-over before marking for review.

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -309,6 +309,7 @@ class MultiField(LayoutObject):
         self.template = kwargs.pop('template', self.template)
         self.field_template = kwargs.pop('field_template', self.field_template)
         self.flat_attrs = flatatt(kwargs)
+        self.has_errors = False
 
     def render(self, form, form_style, context, template_pack=TEMPLATE_PACK, **kwargs):
         # If a field within MultiField contains errors
@@ -316,6 +317,7 @@ class MultiField(LayoutObject):
             for field in map(lambda pointer: pointer[1], self.get_field_names()):
                 if field in form.errors:
                     self.css_class += " error"
+                    self.has_errors = True
 
         field_template = self.field_template % template_pack
         fields_output = self.get_rendered_fields(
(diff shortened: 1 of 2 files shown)


smithdc1:
hi @mulhollandms 

Could I suggest a couple of other options to try fix this which don't need `has_errors`.

1. Change this line to ` {% if form_show_errors and field.errors %}`
https://github.com/django-crispy-forms/django-crispy-forms/blob/77f4ea7edd3d8a4cd21f20bbe7fbce1511248e10/crispy_forms/templates/bootstrap3/layout/multifield.html#L5

2. move `<div class="alert alert-danger" role="alert">` within the for loop, probably after this line? 
https://github.com/django-crispy-forms/django-crispy-forms/blob/77f4ea7edd3d8a4cd21f20bbe7fbce1511248e10/crispy_forms/templates/bootstrap3/layout/multifield.html#L8

These seem to help, but I'm less familiar with bootstrap3

mulhollandms:
@smithdc1 ,

Moving the alert div inside the for loop would mean that if multiple errors are present, each field with errors will have its own separate alert with a list of errors, which looks strange and doesn't seem to line up well with the [documented use](https://github.com/django-crispy-forms/django-crispy-forms/blob/master/docs/layouts.rst#fundamentals):
> this renders all the fields wrapped in a div and when there are errors in the form submission, they are shown in a list instead of each one surrounding the field

updated example:

image


Also, yeah I was a little hesitant about just adding `has_errors` to MultiField but the only other attr that indicates whether this subset of fields has errors is the `css_class`. On that note, I also found that with multiple errors, the MultiField div ends up with `class="ctrlHolder error error"`, which is due to this loop. https://github.com/django-crispy-forms/django-crispy-forms/blob/77f4ea7edd3d8a4cd21f20bbe7fbce1511248e10/crispy_forms/layout.py#L315-L318
Might be worth adding a break to get out of that loop once the error class is added

PR #988, opened 2020-02-20
Flake8 on pull request

smithdc1:
We started to use github actions at the end of last year. The first step is to get flake8 working and currently it is configured to run on push so that it doesn't give errors on pull requests. 

This pull requests finalises flake8 compatibility and therefore I also propose we move to running flake8 for all pull requests. Yes we have #noqa flags in a few places but I think there is value in adding flake8 on `pull_request` now. 

I'll look to raise the current noqa flags on an issue so we can tackle those separately. Some of these (e.g. C901 below) would imply re-write of the main render function and therefore warrants deeper thought.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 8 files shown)


smithdc1:
Ok, this one is merged now as well. 👍

PR #999, opened 2020-02-29
Improve test coverage

smithdc1:
This pull request adds a number of tests for areas of the project which currently don't have test coverage (e.g. formactions/ uneditable field / file field)

I've also made a few amendments to some `if` statements which are not needed and so have removed these. 

The final change I made was to `helper.py` to raise an error if somehow _form_style is incorrectly set. Complete, with a test of course. 😄 

By my workings this should get a small increase in test coverage from 97% to 99%.

### DIFF
--- a/crispy_forms/helper.py
+++ b/crispy_forms/helper.py
@@ -253,8 +253,13 @@ class FormHelper(DynamicLayoutHandler):
         if self._form_style == "default":
             return ""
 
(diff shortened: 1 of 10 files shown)


smithdc1:
I've realised this is impossible to review. Sorry.

I'll split into smaller chunks.

PR #1001, opened 2020-03-03
Run tests with django_coverage_plugin

smithdc1:
I came across django_coverage_plugin and thought it would be interesting to see it applied to this project. 

It's interesting in that it shows we have really low coverage for the templates, particularly the bootstrap ones. If we merge, would we then look to increase coverage in these areas - be it removal of unused templates, or adding new tests. 

I'm expecting coverage to drop in to the 80% range. Let's see if and how codecov responds as I've re-written how travis runs the tests. 

Thoughts?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[run]
+plugins =
+    django_coverage_plugin
+source =
+    crispy_forms
+omit =
+    */.tox/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,13 @@ python:
 cache: pip
 
 install:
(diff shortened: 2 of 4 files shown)


smithdc1:
I've reviewed the coverage report in more some more detail and have a few detailed notes which need investigation. However, I don't think they affects the wider question of 'is this a good idea?'. 

- We're no longer getting branch reporting (--branch flag required?)
- Something odd is happening with crispy_forms_utils. I wonder if we're no longer getting reports appending across runs (-a flag is required?)

smithdc1:
We're dropping the template packs, so this is no longer relevant to this package.

PR #1003, opened 2020-03-08
Remove bs3 styles from crispy_forms_field

smithdc1:
Here is another step towards decoupling _crispy-forms-core_ from template packs. 

This removes the bs3 styles which are currently being added in `crispy_forms_field.py`. 

Interestingly, and at least from what I can see in the docs, `form-control-danger` is not the right way of adding errors in bs3. 'has-error' is the solution and that's already built into the templates. This made it much easier to put the remaining logic into the templates. I suspect the bs4 templates will not be quite so easy.

Here are before/after images (no change hopefully) 🤞 

Before:
before

After: 
after


smithdc1:
@carltongibson Appreciate your thoughts on this. It's another step towards being able to split out the template packs.

PR #1004, opened 2020-03-09
Rendering ClearableFileInput

bryan-brancotte:
Parsing the widget to get the initial data, and rending the current value of the ClearableFileField.

Closing #918


bryan-brancotte:
Hi @smithdc1 

I tried to adresse to issue of ClearableFileField, it works, but I don't like what I have for now.

I tested it with project that use it, but also tested with the MWE of https://github.com/django-crispy-forms/django-crispy-forms/issues/918#issuecomment-545382695 which have a model attribut which is a file. All seems to work.

There is still some aspect that I don't like, the main one being that I have to loop over the subwidgets, I find it dirty. Also I don't find the solution very beautiful, visually but also with the "css hack" I had to use to make the clear checkbox float on the right.

smithdc1:
Hi @bryan-brancotte 

Thanks for looking into this. I've been thinking about this over the last day and here are some initial thoughts. 

Django has a template for [clearable_file_input](https://github.com/django/django/blob/master/django/forms/templates/django/forms/widgets/clearable_file_input.html). I was hoping that this could give some ideas on how to tackle this problem but as you have pointed out we need to get the data from the subwidget. i.e. the name fields are on `field.subwidgets...` not `field.field.widget`. 

The [django tests](https://github.com/django/django/blob/8d76443aba863b75ad3b1392ca7e1d59bad84dc4/tests/forms_tests/widget_tests/test_clearablefileinput.py) are also helpful, should we try and replicate some of these to ensure we have the same logic. e.g. when `is_required` is `True` or `False` etc. 

My final thought for today is on the checkbox. I think what you have done with the checkbox is impressive to get it to look like that. However, I've got a few thoughts, not quite sure what the answer is as yet. A quick google on the topic wasn't very successful, I ended straight back here 🤣 

- Is it obvious to an end user how to use it? (this combines checkbox + file, Django default splits them out)
- I think currently there is quite a different visual look depending if there is an initial value?
- I think one of the questions is about displaying an initial value for a the default file field (non clearable). Have we given a view on this?

bryan-brancotte:
Hi David

Indeed tests are helpful but form now I have completly neglected them, we should add some.

I agree that the widget in itself is not very understandable, but so does the default django one. Here is what django doesand what we have for now :

Capture d’écran de 2020-03-09 10-13-03

Capture d’écran de 2020-03-11 09-04-18

Maybe adding on the left of the second line a "Change" in order to indicate that the new value is here ? The "Browse" is an ::after element, so the change would have to be a ::before but could be achieved with a [scoped css](https://vue-loader.vuejs.org/guide/scoped-css.html)

bryan-brancotte:
With the "Change" on the second line:
Capture d’écran de 2020-03-11 10-09-38

smithdc1:
Hi Bryan,

The image in your latest post looks great. I think it's much clearer to the end user what going on. 

A couple of questions (sorry, still learning lots)

This works great when it's just one file. What happens when the requirement is for multiple files?

If form is submitted and fails, would the widget change from the standard file widget to the clearable widget? Is that ok?

David

bryan-brancotte:
Hi David, 

Question are not silly, just unanswered.

That I know of there is no multiple FileField in models, so the widget is ok. When considering FileField in a Form, we can set multiple to true, and in this case the input type=file handle multiple file (not tested now, but few months back)

When the submission fails, that I remember, the new file is lost. For me it's common behavior, not normal, but common.

Bryan

smithdc1:
@bryan-brancotte 
Super. Time for docs/tests, I think. 

Anything I can help with?

bryan-brancotte:
Good idea @smithdc1 , I don't know if we have test for filefield from ModelFrom, maybe writting some ? I will be less active in the following weeks as long as we are under (partial) lockdown in France

bryan-brancotte:
Capture d’écran de 2020-03-19 12-17-16
the help text is overflowing in horizontal form (in test project you have to comment `helper.use_custom_control = False` to see it)

smithdc1:
Hi @bryan-brancotte hope you are well.
I've added documentation, another small step.

smithdc1:
I've now rebased for the merge conflict. A little tricky this one as it was in the middle of other changes, so I just added the small amend as another commit.

smithdc1:
Hi @bryan-brancotte, I've added some tests now as well.

If you could review my docs / tests I'd appreciate it. I'll leave some comments to try and explain my thinking - a little unsure when it's appropriate to leave a comment in the code! Think I think we're good to squash and merge 🚀 

Once this is merged do you think we should release 1.9.1? We've merged in a handful of small changes over the last quarter.

crispy_forms/tests/test_layout.py
```diff
@@ -598,3 +599,54 @@ def test_update_attributes_class():
...
     assert html.count(' class="hello hello2 textinput') == 1
+
+
+@only_bootstrap4
+def test_file_field():
+    form = FileForm()
+    form.helper = FormHelper()
+    form.helper.layout = Layout("clearable_file")
+    html = render_crispy_form(form)
+    assert contains_partial(
```
smithdc1, line 610: We don't need to worry about whitespace with contains_partial. I did try using Django's `assertinhtml` but ran into trouble with this. I think it's a unit test thing 🤷‍♂ . Anyway, we have contains partial which is similar.

crispy_forms/tests/test_layout.py
```diff
@@ -598,3 +599,54 @@ def test_update_attributes_class():
...
+        <span class="input-group-text">Currently</span></div><div class="form-control d-flex">
+        <span style="flex-grow: 1;"> <a href="something">something</a> </span> <span class=" ">
+        <span class="custom-control custom-checkbox">
+        <input type="checkbox" name="clearable_file-clear" id="clearable_file-clear_id" class="custom-control-input">
+        <label class="custom-control-label mb-0" for="clearable_file-clear_id">Clear</label></span></span></div></div>
+        <div class="input-group mb-0"><div class="input-group-prepend"> <span class="input-group-text">Change</span>
+        </div><div class="form-control custom-file" style="border:0">
+        <input type="file" name="clearable_file" class="custom-file-input">
+        <label class="custom-file-label" for="id_clearable_file">---</label></div></div>
+        <div class="input-group mb-0"> </div></div>""",
```
smithdc1, line 622: I've run a test for the 4 options for file fields. Standard + clearable and Django + Boostrap. 
This test is a little long... but was a little bit unsure where it could be shortened.

crispy_forms/tests/forms.py
```diff
@@ -144,3 +144,20 @@ class SampleForm8(forms.ModelForm):
     class Meta:
         model = CrispyTestModel
         fields = ("email", "password2", "password")
+
+
+class FakeFieldFile:
```
smithdc1, line 149: FakeFieldFile is a direct copy from Django's own file field tests.

smithdc1:
Final comment, where did TravisCI go? 
I know I've opened a PR which proposes to drop it, but it hasn't run against this PR. (At time of writing this comment)

bryan-brancotte:
Hi David,

Sorry for being too silent, thanks for the docs and tests ! For the docs nothing to add.

I took a look at the test and found that contains_partial does not look at the children of the needle (the ```<script></script>``` element was missing but the test still passing which was surprising). I updated the function and test to highlight that is only search for one element. Thus for now test are failing as the function is used in test_file_field.

I think we can rework the test to ensure that each element are here when needed such as ```<span class="custom-control custom-checkbox">``` only when use_custom_control is True. Regarding the order of the elements, I don't have an out of the box solution for now, I keep looking for it.

See you soon

Bryan

smithdc1:
Hi @bryan-brancotte 
Hope you and your family are well. 

Thanks for reviewing the tests - I've simplified to only check they correct type of widget renders.

I'll have a final review (later hopefully) and then look to merge! 🚀

cpina:
> Hi @bryan-brancotte
> Hope you and your family are well.
> 
> Thanks for reviewing the tests - I've simplified to only check they correct type of widget renders.
> 
> I'll have a final review (later hopefully) and then look to merge! rocket

I just want to say thanks for this - I'm really looking forward to having it! I'll test it soon. The original field confused some of our users and I wanted to look at how to improve it.

smithdc1:
Hi all

I've had a final look over and this looks good to me, so will now merge. 

@cpina - I agree, Bryan has done a great job with this. 👏 Just one note re your PR on Django - I don't think this design will work particularly well with a disabled file field? 🤔

cpina:
> Hi all
> 
> I've had a final look over and this looks good to me, so will now merge.
> 
> @cpina - I agree, Bryan has done a great job with this. clap Just one note re your PR on Django - I don't think this design will work particularly well with a disabled file field? thinking

I've had a look. I'm sure that you know but for reference for me later/others: with the Django PR https://github.com/django/django/pull/12864 (not merged) and django-crispy-forms with and self.helper.use_custom_control = False: the disabled works fine (with the Django original widget obviously)

With the custom Django Crispy Form widget: the disabled has no effect (to the checkbox or the Browse button).

I'm checking the new Django Crispy Form out of curiosity where this would be implemented...

cpina:
Two quick things (and I'm happy to open a ticket). In one of my projects where I don't use very nice file names this is what the new ClearableFileInput renders:
Screenshot_20200509_115350
Would you like that I open an issue?

Regardless of this: the new widget looks way better!

Also, for the `disabled` attribute related to this I see that `venv/src/django-crispy-forms/crispy_forms/templates/bootstrap4/layout/field_file.html` would need to do something similar as the Django https://github.com/django/django/pull/12864 . But doing it is not enough because `venv/src/django-crispy-forms/crispy_forms/utils.py` render_field function is not passing the `disabled` into `template.render`. It has access via `field_instance.disabled`. I don't know if it's the best place for this logic as I'm a bit lost there, first time today that I'm looking at django-crispy-forms internal code properly. I thought of something similar to the `hidden` but `attrs` is None so it's not going there. I'm happy to have another look some time soon.

smithdc1:
@cpina yes , please open an issue (or two) so we can track these. 

On the rendering issue do you have any ideas on what should happen from a UX point of view. If you have any ideas on how to implement that suggestion that would also be very much welcomed.

cpina:
> @cpina yes , please open an issue (or two) so we can track these.
> 
> On the rendering issue do you have any ideas on what should happen from a UX point of view. If you have any ideas on how to implement that suggestion that would also be very much welcomed.

Done:
https://github.com/django-crispy-forms/django-crispy-forms/issues/1027
https://github.com/django-crispy-forms/django-crispy-forms/issues/1028

smithdc1:
Not a problem, Thanks so much for all of your input!

PR #1008, opened 2020-03-19
Improve pattern to encompass BS4 css col class such as col-4

bryan-brancotte:
When using horizontal form like in crispy test project, allow to use col-4 is not only col-sm-4:

```
class HorizontalMessageForm(forms.Form):
    (...)

    helper = FormHelper()
    (...)
    helper.form_class = 'form-horizontal'
    helper.label_class = 'col-4'
    helper.field_class = 'col-8'
```

Before:
Capture d’écran de 2020-03-19 15-34-43
Fixe:
Capture d’écran de 2020-03-19 15-35-03

### DIFF
--- a/crispy_forms/helper.py
+++ b/crispy_forms/helper.py
@@ -1,12 +1,11 @@
 import re
 
-from django.urls import NoReverseMatch, reverse
-from django.utils.safestring import mark_safe
-
 from crispy_forms.exceptions import FormHelpersException
 from crispy_forms.layout import Layout
 from crispy_forms.layout_slice import LayoutSlice
 from crispy_forms.utils import TEMPLATE_PACK, flatatt, list_difference, render_field
+from django.urls import NoReverseMatch, reverse
+from django.utils.safestring import mark_safe
 
 
 class DynamicLayoutHandler:
@@ -337,10 +336,10 @@ class FormHelper(DynamicLayoutHandler):
 
         if template_pack == "bootstrap4":
             if "form-horizontal" in self.form_class.split():
-                bootstrap_size_match = re.findall(r"col-(xl|lg|md|sm)-(\d+)", self.label_class)
+                bootstrap_size_match = re.findall(r"col(-(xl|lg|md|sm))?-(\d+)", self.label_class)
                 if bootstrap_size_match:
-                    offset_pattern = "offset-%s-%s"
-                    items["bootstrap_checkbox_offsets"] = [offset_pattern % m for m in bootstrap_size_match]
+                    offset_pattern = "offset%s-%s"
+                    items["bootstrap_checkbox_offsets"] = [offset_pattern % (m[0], m[-1]) for m in bootstrap_size_match]
         else:
             bootstrap_size_match = re.findall(r"col-(lg|md|sm|xs)-(\d+)", self.label_class)
             if bootstrap_size_match:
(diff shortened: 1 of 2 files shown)


crispy_forms/helper.py
```diff
@@ -337,10 +337,12 @@ def get_attributes(self, template_pack=TEMPLATE_PACK):  # noqa: C901
 
         if template_pack == "bootstrap4":
```
smithdc1: What do you think about adding this whole section to a new function? Would it make helper simpler, and this piece easier to test?

bryan-brancotte: I am not sure that it is necessary: I extended the test checking the column are well transmitted for BS4 with 7ac2f7f. Test with BS3 have already been written [here](https://github.com/django-crispy-forms/django-crispy-forms/pull/1008/files#diff-46a067d22cfed9b9ab34736364f083c5R788-R816).

bryan-brancotte:
And thanks for fixing isort David !

crispy_forms/tests/test_form_helper.py
```diff
@@ -870,6 +873,16 @@ def test_label_class_and_field_class_bs4_offset_when_horizontal():
     assert '<div class="offset-sm-3 offset-md-4 col-sm-8 col-md-6">' in html
     assert html.count("col-sm-8") == 7
 
+    # Test col-X pattern
+
+    form.helper.label_class = "col-5 col-lg-4"
+    form.helper.field_class = "col-7 col-lg-8"
+    html = render_crispy_form(form)
```
smithdc1: This tests both col-xx-yy and col-x patterns. Could we combine test here (row 878-885) with row (865-874)?

crispy_forms/tests/test_form_helper.py
```diff
@@ -870,6 +873,16 @@ def test_label_class_and_field_class_bs4_offset_when_horizontal():
...
+
+    assert '<div class="form-group row">' in html
+    assert '<div class="offset-5 offset-lg-4 col-7 col-lg-8">' in html
+    assert html.count("col-7") == 7
```
smithdc1: I wonder if we could do with a SampleForm with just one input in it. Is it important, in this case, that we're testing that we get the same answer for many different inputs (7 in this case, or is it?!) ? 

I've been pondering for a while about how we write some of these tests. I wonder if we can generate small enough html we can 'just' dump the whole thing into the test. It's then clearer what we're actually testing for. At the moment every time I look at these, I jump into the debugger to evaluate html. 

But maybe I'm doing it all wrong 🤷‍♂ (I don't know what I don't know....)

crispy_forms/helper.py
smithdc1: ok 👍

smithdc1:
Hi Bryan,

A couple of comments. I also don't _think_ this needs anything in the docs.

Happy to squash and merge if you are. 👍

crispy_forms/tests/test_form_helper.py
bryan-brancotte: Dumping the whole HTML in the test would be very sensitive to even a new blank space inserted. But in the meantime we could load the html string as an html object, and then dump it. It would be more robust solution to prevent regression but it would also take time to re-write all the test :-/

bryan-brancotte:
Hi @smithdc1 If you are ok with the new tests I wrote, feel free to squash and merge ;o)

smithdc1:
Hi @bryan-brancotte Thanks!

bryan-brancotte:
You are welcome!

crispy_forms/tests/test_form_helper.py
smithdc1: Thanks for the insight 👍

PR #1026, opened 2020-05-09
Release 1.9.1

smithdc1:
Hi all

I propose we issue a new release now the new `ClearableFileInput` rendering has been merged. 

The change list is [here](https://github.com/django-crispy-forms/django-crispy-forms/milestone/10?closed=1), in addition to the new file field there are a few small bug fixes that have been merged since the last release. 

If there is agreement I'll look to issue over the coming days?

(p.s. little unsure if this should be 1.9.1 or 1.10.0)


smithdc1:
We probably want to take a look at #1028 before this. I _think_ it needs some JS??

#1027 I'm slightly less concerned about as it's not imediately obvious what the solution is (how would a disabled bootstrap 4 file field look anyway?!)

smithdc1:
> We probably want to take a look at #1028 before this. I _think_ it needs some JS??
> 
> #1027 I'm slightly less concerned about as it's not imediately obvious what the solution is (how would a disabled bootstrap 4 file field look anyway?!)

I've been looking into this issue this afternoon. Other examples of bootstrap file boxes all have same issue, and is an issue with the file field that's already in 1.9.0. I'll therefore look to issue the release, tentively target next weekend. 

[link1](https://mdbootstrap.com/docs/jquery/forms/file-input/)
[link2](https://www.w3schools.com/bootstrap4/bootstrap_forms_custom.asp)

cpina:
Hi,

> > We probably want to take a look at #1028 before this. I _think_ it needs some JS??
> > #1027 I'm slightly less concerned about as it's not imediately obvious what the solution is (how would a disabled bootstrap 4 file field look anyway?!)
> 
> I've been looking into this issue this afternoon. Other examples of bootstrap file boxes all have same issue, and is an issue with the file field that's already in 1.9.0. I'll therefore look to issue the release, tentively target next weekend.

In the issue https://github.com/django-crispy-forms/django-crispy-forms/issues/1027 I suggested disabling the checkbox and the Browse button (second screenshot). Do you think that this is a possible solution? It avoids the user thinking that the file can be changed. I thought (but I'll check) that the new file field is just a combination of a checkbox+label+button, etc. (like the original Django).

Either way this week I'll use master of django-crispy-forms at least in my development environment.

cpina:
> Hi,
> 
> > > We probably want to take a look at #1028 before this. I _think_ it needs some JS??
> > > #1027 I'm slightly less concerned about as it's not imediately obvious what the solution is (how would a disabled bootstrap 4 file field look anyway?!)
> > 
> > 
> > I've been looking into this issue this afternoon. Other examples of bootstrap file boxes all have same issue, and is an issue with the file field that's already in 1.9.0. I'll therefore look to issue the release, tentively target next weekend.
> 
> In the issue #1027 I suggested disabling the checkbox and the Browse button (second screenshot). Do you think that this is a possible solution? It avoids the user thinking that the file can be changed. I thought (but I'll check) that the new file field is just a combination of a checkbox+label+button, etc. (like the original Django).
> 
> Either way this week I'll use master of django-crispy-forms at least in my development environment.

E.g., in
https://mdbootstrap.com/docs/jquery/forms/file-input/

I can change add disabled="" here:
` <input type="file" class="custom-file-input" id="inputGroupFile01" aria-describedby="inputGroupFileAddon01" disabled="">`

And the "Browse" button then is disabled, would this be a possible solution?

smithdc1:
Thanks all. 🚀

smithdc1:
https://pypi.org/project/django-crispy-forms/1.9.1/

bryan-brancotte:
Many thanks to you David !

PR #1031, opened 2020-05-12
Avoids possible XSS when attaching files.

### LINKED ISSUE #1030: ClearableFileInput: possible XSS attack
* Package version: 2ec7b692f948c25917482f0f723b0332acbd809d 
* Django version: 3.0.5
* Python version: 3.7.3
* Template pack: Bootstrap4

### Description:
I was looking at the new clearable file field and its HTML (to adjust some existing code).

In https://github.com/django-crispy-forms/django-crispy-forms/blob/master/crispy_forms/templates/bootstrap4/layout/field_file.html#L38 it's using innerHTML function: this can create an XSS attack to someone.

To reproduce, create a file with a name like: ``. I've done it using the Python console:
```
f = open("""<img src=x onerror="alert(\'XSS Attack\')">""", "w")
f.write('test')
f.close()
```

Then attach this file in a field file and see "XSS Attack". See (random Google) this page for explanation: https://gomakethings.com/preventing-cross-site-scripting-attacks-when-using-innerhtml-in-vanilla-javascript/

The fix is easy: replace `innerHTML` by `textContent`. I can't think (in this case) what `innerHTML` would do that `textContent` doesn't.

cpina:
Closes #1030

### DIFF
--- a/crispy_forms/templates/bootstrap4/layout/field_file.html
+++ b/crispy_forms/templates/bootstrap4/layout/field_file.html
@@ -35,7 +35,7 @@
                 for (let i=0;i<e.target.files.length;i++){
                     filenames+=(i>0?", ":"")+e.target.files[i].name;
                 }
-                e.target.parentNode.querySelector('.custom-file-label').innerHTML=filenames;    
+                e.target.parentNode.querySelector('.custom-file-label').textContent=filenames;    
             }        
         </script>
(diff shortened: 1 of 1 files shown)


cpina:
> nothing to add, thanks for the fix !

Thanks to you for the PR, the file widget looks much much better!

PR #1040, opened 2020-06-19
Changed urlpatterns to use path

smithdc1:
url() is deprecated in Django4.0


smithdc1:
Thanks. It has amused me immensely that I've needed to change this.

PR #1041, opened 2020-06-20
1.9.2 release

smithdc1:
* Fixed FileField UI bug

@bryan-brancotte @carltongibson 

I _think_ the `FileField` bugs have finally been fixed (🥳 ) with great work by @cpina / @bryan-brancotte . It's a small release but I think it's worth sharing as we don't have many other changes due imminently (there's some more recent issues, but I've not even started to look at fixing them).

If your happy, I'll publish a release.

David


cpina:
I've been using the new, fixed FileField all the time the last days and I really like it.

I wonder only one thing and it's if the "Clear" checkbox should be centered or not when there are multiple lines. Currently in master we have:
current-01
current-02

When I saw it (in my application but it's the same case) I wondered if it should look like:
current-03
current-04

I have no strong opinion... perhaps I'm slightly inclined for the centered one. I've done (having in mind, again, backend dev playing with CSS here!):
current-05
I haven't checked if the normal case (shorter current file) keeps working fine.
(I've only added `align-self-center`)

Any opinion on this one?

smithdc1:
Ok, let's have a look at this. We can publish a release once we're happy. Thanks for the report!

I've opened a new issue to discuss this. #1043 

David

PR #1047, opened 2020-06-24
Fixed alignment of checkbox in filefield

### LINKED ISSUE #1043: FileField Clear Checkbox Alignment
Below is copied from #1041 by @cpina 

I've been using the new, fixed FileField all the time the last days and I really like it.

I wonder only one thing and it's if the "Clear" checkbox should be centered or not when there are multiple lines. Currently in master we have:
current-01
current-02

When I saw it (in my application but it's the same case) I wondered if it should look like:
current-03
current-04

I have no strong opinion... perhaps I'm slightly inclined for the centered one. I've done (having in mind, again, backend dev playing with CSS here!):
current-05
I haven't checked if the normal case (shorter current file) keeps working fine.
(I've only added `align-self-center`)

Any opinion on this one?

smithdc1:
@bryan-brancotte @cpina 

Appreciate thoughts. 

Thanks both. 

Fixes #1043 

**Before**
image

**After**
image

### DIFF
--- a/crispy_forms/templates/bootstrap4/layout/field_file.html
+++ b/crispy_forms/templates/bootstrap4/layout/field_file.html
@@ -12,7 +12,7 @@
         <a href="{{ field.value.url }}">{{ field.value }}</a>
         </span>
         {% if not widget.data.required %}
-        <span class=" ">
+        <span class="align-self-center ml-2">
             <span class="custom-control custom-checkbox">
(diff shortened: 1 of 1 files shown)


cpina:
Tested in the project that I'm working: it looks perfect! :+1: 
With this merged I'd be happy releasing 1.9.2 as we had mentioned last week.

smithdc1:
Ok thanks all

PR #1052, opened 2020-07-05
Suppressed flake8 error

smithdc1:
New `flake8` version starts to raise an error for not having imports at the top of the file (E402)

Tests fail if register variable is not created before importing filters


smithdc1:
Failures are unrelated (Codecov timeout)

PR #1053, opened 2020-07-05
Make flake8 fail loudly

smithdc1:
Flake8 github action is currently failing silently. 

If I remember correctly this was so we could get it working and then go and fix and errors. Now that we have it working let's make it fail loudly. 

This SHOULD fail as the new version of flake8 raises an error see #1052.

Once that is merged, we can re-run the checks on this PR and it should then pass.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -17,18 +17,7 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


smithdc1:
https://github.com/django-crispy-forms/django-crispy-forms/pull/1053/checks?check_run_id=838590439

Ok, this is working as expected.

cpina:
This seems ready to be merged but the logs for the run have been purged. Do you want to rebase and re-run a build?

PR #1055, opened 2020-07-08
Doc'd form-inline class needs to be added to <form>

smithdc1:


jerinpetergeorge:
I would like to add a simple example too, because
> _"Talk is cheap. Show me the code"_

<pre><code>{% load crispy_forms_tags %}

{% block form_body %}
    <strike>&lt;form method="post"></strike>
        {% crispy form %}
    <strike>&lt;/form></strike>
{% endblock %}</code></pre>

smithdc1:
Hi - thank you so much for the feedback, it is really appreciated. 

Personally, I'm _not sure_ about this. Reading from the top of the page it's explained there are no `<form>` tags needed in the template (but there is some caveats when it may be useful to do it yourself), so I don't think we need a full example here. 

Let's see if we get a 3rd opinion on this before merging.

jerinpetergeorge:
Tha's kind of true:laughing: Let's see other's opinion about this.

cpina:
> Hi - thank you so much for the feedback, it is really appreciated.
> 
> Personally, I'm _not sure_ about this. Reading from the top of the page it's explained there are no `<form>` tags needed in the template (but there is some caveats when it may be useful to do it yourself), so I don't think we need a full example here.
> 
> Let's see if we get a 3rd opinion on this before merging.

Are we discussing this commit? https://github.com/django-crispy-forms/django-crispy-forms/pull/1055/commits/b5fc8eb3e5ae34348c4b22e91a71c5c6180c82a0

My opinion: :+1: 

I haven't used `form-inline` neither tested them nwo. But this commit is in a Bootstrap3 section but I understand that it would also apply to Bootstrap4. Do you know if it apply on Bootstrap4? Should it be somewhere else? I can check a bit more if unsure.

cpina:
I've commented in the issue but just a quick copy-paste:

Side note: It seems that `form-inline` is removed from Bootstrap5 (see https://getbootstrap.com/docs/5.0/getting-started/download/ section "Class Updates")

PR #1068, opened 2020-10-10
Added Python 3.9 to test matrix 

smithdc1:


smithdc1:
Looks like the action needs updating. (Fingers crossed 🤞)

smithdc1:
So this now works, but is failing on master due to an unrelated issue. 

Looks like GitHub actions does not have a nice way of "allow failures". More investigation required...

PR #1074, opened 2020-11-02
Split Development requirements into separate file

smithdc1:
We don't need to install all of the requirements to run the test suite?

Let's see what impact it has on the test run times.


smithdc1:
GH actions doesn't seem to run any faster with this reliably (it should be about 40seconds quicker). Other factors seem to vary more wildly, such as how long it takes to restore the pip cache.

PR #1080, opened 2020-11-08
CSS class for input group size in Bootstrap 4

medihack:
There is no `input-sm` and `input-lg` CSS class in Bootstrap 4. It uses several other classes to adjust the size in form controls (`form-control-sm`, `form-control-lg`, `custom-select-sm`, `custom-select-lg`, `input-group-sm`, `input-group-lg`). For input groups it is `input-group-sm` and `input-group-lg` (but together with the `input-group` class)`.

Fixes #1071

PR #1085, opened 2020-11-17
Release 1.10

smithdc1:
Prep for next release. 

There is one more PR I'd like to get into this if we can (#1080), but there is no rush. I can easily prep another release as there is plenty here. 

It's surprising that even with something as stable as `crispy-forms` we've had more than a dozen merged pull requests since the last release. Yes many are small, but all improvements to help move it forward. 

@carltongibson anything to add? I was thinking of pushing this to publish over the weekend (time & family needs depending).


carltongibson (approved):
Looks good. Super work. (My approach would be to get it out — no point having inventory on the shelf…)

Thanks @smithdc1!

smithdc1:
Ok -- I've started to have a look at publishing this. Need to get setup on the pi. I'm **so** missing git-fork. Any alternatives you would recommend?

I've tried vscode extensions, but I'm missing the simple squash / rename functionality from git-fork.

smithdc1:
https://pypi.org/project/django-crispy-forms/1.10.0/  :tada:

PR #1091, opened 2020-12-11
Custom select

cpina:
Add `class="custom-control custom-select"` for the select widgets.

Before (see the arrow at the end):
select

After (Bootstrap4 arrow):
custom_select

When it has been seen it cannot be unseen!

In django-crispy-forms Bootstrap4 there is no specific template for the select widget: it relies on the Django template. This is good (one less think to maintain) but conceptually I think that would prefer to avoid having Bootstrap4/template-pack specific code in `crispy_forms_field.py`: "custom-control custom-select" could be moved into a new template for "select" like it's done for other widgets. On the other hand it's handy to not have the template and piiggyback Django select (`django/forms/templates/django/forms/widgets/select.html`) (other possible refactorings would be possible). Perhaps when working on Bootstrap5 it would be a good moment to re-consider this.

The Bootstrap4 (or Bootstrap3) code in `crispy_forms_field.py` dates from 2013.


crispy_forms/templatetags/crispy_forms_field.py
```diff
@@ -130,6 +130,10 @@ def render(self, context):  # noqa: C901
                     css_class += " form-control-danger"
 
             if template_pack == "bootstrap4" and not is_multivalue(field):
+                if is_select(field):
```
smithdc1, line 133: We're trying to push all of the logic into the templates rather than code. (It'll make it easier to extract templates from core).

You can pass in the additional classes into the django rendered widget via the template. See the BS5 `field` template for some inspiration as I've done the errors using this technique. This avoids needing the standalone template then html layout is the same and it's just the classes that change. 

(I'm on mobile at the moment but can draft this up at some point if you need help).

cpina: The separation between templates code and Python was my concern as well. I'll have a look at BS5 `field` template and I have an unorthodox solution that I'd like to try as well.

cpina: I have a solution but I'm not 100% that it's what you meant (and I have a question with the preappend/postappend widgets). I'll push the code in a few hours when I'm back home :-)

cpina:
With the latest commit (12th Dec): the logic to fix this is in the templates.

This has a problem: the custom-select `select` does not work for Prepended/Appended texts. This is not only for the `select` but for other widgets. See for more details  #1093 .

So this case works fine:
select-good

But this case is not working yet:
select-bad

But since this is related to any widget I would plan a solution in #1093

crispy_forms/templatetags/crispy_forms_field.py
cpina: > You can pass in the additional classes into the django rendered widget via the template. See the BS5 `field` template for some inspiration as I've done the errors using this technique. This avoids needing the standalone template then html layout is the same and it's just the classes that change.

@smithdc1 is this what you were thinking?
https://github.com/django-crispy-forms/django-crispy-forms/pull/1091/commits/01405272f3b833ba403b945d62ae4b7d83a6aa9b

If not: let me know, not a problem at all!

crispy_forms/templatetags/crispy_forms_field.py
```diff
@@ -27,6 +27,11 @@ def is_select(field):
     return isinstance(field.field.widget, forms.Select)
 
 
+@register.filter
+def is_selectsingle(field):
+    return is_select(field) and not field.field.widget.allow_multiple_selected
```
smithdc1, line 32: Why do we need to limit this to just single selects? The multiple select also supports `custom-select`? From my testing I can't see any difference in the styling on multiple selects between `custom-select` and `form-control`

crispy_forms/templates/bootstrap4/field.html
```diff
@@ -37,6 +37,11 @@
                 {% include 'bootstrap4/layout/help_text_and_errors.html' %}
             {% elif field|is_file and use_custom_control  %}
                 {% include 'bootstrap4/layout/field_file.html' %}
+            {% elif field|is_selectsingle and use_custom_control %}
```
smithdc1, line 40: I did wonder about moving this down inside the next div (I think it works being the same number of lines)

crispy_forms/templates/bootstrap4/field.html
```diff
@@ -37,6 +37,11 @@
...
+                <div class="{{ field_class }}">
+                     {% crispy_field field 'class' 'custom-control custom-select' %}
```
smithdc1, line 42: I think we need `custom-select` works just fine on it's own?

crispy_forms/tests/test_layout.py
```diff
@@ -606,6 +607,34 @@ def test_use_custom_control_is_used():
     assert response.content.count(b"custom-checkbox") == 0
 
 
+@only_bootstrap4
```
smithdc1, line 610: Could this benefit from `@pytest.mark.parametrize`? 

https://docs.pytest.org/en/stable/parametrize.html

docs/form_helper.rst
```diff
@@ -159,23 +165,17 @@ All previous, ``bootstrap`` (version 2 and 3) attributes are also settable in bo
 .. |customfile| image:: images/custom_file_field.png
    :width: 300px
    :align: middle
-.. |clearablefile| image:: images/clearable_file.png
```
smithdc1, line 162: Dropping clearable file from here is an unrelated change?

docs/form_helper.rst
```diff
@@ -135,22 +135,28 @@ Bootstrap 4 Helper attributes
...
    :align: middle
 .. |customcheckbox| image:: images/custom_checkbox.png
    :width: 300px
    :align: middle
-.. |radio| image:: images/radio.png    
+.. |select| image:: images/select.png
    :width: 300px
    :align: middle
-.. |customradio| image:: images/custom_radio.png
+.. |customselect| image:: images/custom_select.png
```
smithdc1, line 159:
```
/home/pi/Projects/django-crispy-forms/docs/form_helper.rst:156: WARNING: image file not readable: images/select.png
/home/pi/Projects/django-crispy-forms/docs/form_helper.rst:159: WARNING: image file not readable: images/custom_select.png
```

I can't seem to build the docs?

smithdc1:
Hi Carles, thanks for looking at this, this is a nice improvement. I've left a few small comments. 

I think we should take this as a standalone item which covers the most likely use case. We can then look at a select within a append / prepended wrapper as a separate item? 

I'd be happy to push a release with just this change.

crispy_forms/templatetags/crispy_forms_field.py
smithdc1: Yes, like this. :+1:

docs/form_helper.rst
cpina: Yes (I noted this in the commit, perhaps I should move it to a PR): https://github.com/cpina/django-crispy-forms/commit/c1340236525c5712f05184488ccc45580f77a8d9

The file field custom clearable png is, at the moment, broken in the documentation: https://django-crispy-forms.readthedocs.io/en/latest/form_helper.html?bootstrap-4-helper-attributes#bootstrap-4-helper-attributes

crispy_forms/templates/bootstrap4/field.html
cpina: I'm changing it, you are right: https://getbootstrap.com/docs/4.5/components/forms/#select-menu

The majority of previous controls like the `custom-switch` also have `custom-control`, this might have confused me :-)

crispy_forms/templatetags/crispy_forms_field.py
cpina: I didn't see any difference either. I thought of minimising the changes in the HTML output (to avoid breaking changes to users). Since Bootstrap documentation says explicitly that it is supported I'm happy to change it an apply let the `custom-select` go to the multiple selects. Actually it might be even less surprising for users. It also have fewer code for us.

It also occurred to me that some other template backend might need this distinction (single vs multiple) but we can cross this bridge when you come to it.

crispy_forms/templates/bootstrap4/field.html
cpina: I've changed it. My thoughts were between having one more nested `if` (after the change) or the similar pattern that we had for other fields.

cpina:
> Hi Carles, thanks for looking at this, this is a nice improvement. I've left a few small comments.
> 
> I think we should take this as a standalone item which covers the most likely use case. We can then look at a select within a append / prepended wrapper as a separate item?
> 
> I'd be happy to push a release with just this change.

I'm also happy to address the pre/post append select issues as part of #1093 . I think that pre/post append might need bigger changes and longer discussion.

docs/form_helper.rst
cpina: I'll amend this (I missed adding some files). How are you building the docs? I can't see in the documentation or in the .github action.

cpina:
```
cd docs
make html
```
I might add this in the docs (or read them better if it's there)

docs/form_helper.rst
cpina: > Dropping clearable file from here is an unrelated change?

I've created a branch minimising this change: https://github.com/cpina/django-crispy-forms/tree/custom-select-with-minimal-docs-change . I might have got a bit carried on the other day. I can make the other changes (delete the clearable one maybe, re-order the table so it matches the text) in another PR.

crispy_forms/templatetags/crispy_forms_field.py
smithdc1: https://docs.djangoproject.com/en/3.1/ref/forms/api/#django.forms.BoundField.widget_type Projects using Django 3.1+ can get a similar ability directly in the templates.

docs/form_helper.rst
smithdc1: The clearable file also looks awful on rtd, I'm not sure it's that helpful.

smithdc1 (approved):
This looks good to me! :tada:

Issue #1095, opened 2020-12-14
Fix a11y errors for bootstrap 4


yuqiuw:
* Package version:
* Django version:
* Python version:
* Template pack: (Optional)

### Description:

This tracks testing Django form components in use by Incubator for accessibility issues and remedying these issues. 

Reference requirements for Django form: https://www.w3.org/TR/wai-aria-practices/examples/landmarks/form.html

### Expected results:

The code should solve all a11y problems detected by Axe and Wave

yuqiuw:
Fixed in https://github.com/django-crispy-forms/django-crispy-forms/pull/1096

smithdc1:
I'm keen to improve accessibility issues in the template packs.

I'll close this for now as this is a generic accessibility request. I'd prefer to have issues targeting specific issues (e.g. https://github.com/django-crispy-forms/crispy-bootstrap5/issues/152) that can be improved. That'll allow easier judgement of when an issue is resolved.

PR #1096, opened 2020-12-14
Fix a11y for bt4

yuqiuw:
https://github.com/django-crispy-forms/django-crispy-forms/issues/1095


smithdc1:
Thanks for the interest here. I'm super supportive of any improvements in this area (although I'm not immediately familiar with the details).

On initial review I'm not sure this is right?

The Mozilla dev docs say labels use `for` not`id`?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

The answer to this SO question says we don't need to use them in this context?

https://stackoverflow.com/questions/11152323/the-purpose-of-using-aria-labelledby-on-already-labeled-input-elements

yuqiuw:
> https://stackoverflow.com/questions/11152323/the-purpose-of-using-aria-labelledby-on-already-labeled-input-elements

Yeah it just switches to use `aria-labelledby` instead of for to support linking multiple input fields with the same label. 

A good example is like this field in a django form
```
# SplitHiddenDateTimeWidget
    splitHiddenDateTimeWidget = forms.SplitDateTimeField(
        widget=forms.widgets.SplitHiddenDateTimeWidget
    )
```

yuqiuw:
Hi @smithdc1, can I get approved and merge this in? Thank you for reviewing!!

smithdc1:
Hi @yuqiuw I'll need some time to think about this. 

As mentioned above the guidance I found days to use `for` rather than the aria attributes?

What am I missing here?

yuqiuw:
Hi @smithdc1 , You're totally right that the `for` shouldn't be exchanged. 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

I put the `for` back, and the `aira-labelledby` only works as a replenish so that a label can label several inputs.

Thank you for looking at this!

yuqiuw:
Before: 
image

After:
image

Also fixes the accessibility problem shown in Axe.

yuqiuw:
Hi @smithdc1 , can I have some updates on this? Thank you so much! Merry Christmas to you and your family!

smithdc1:
Hi I'll have a look at this after the holiday period.

A few questions at this stage. 

- I think you are suggesting this is only an issue for items where there are multiple inputs for a single label (e.g split date time). A single text back is currently ok?

- This is limited to a single template pack but presumably this applies to all template packs?

- As a principle were trying to put template pack specific logic into the templates.

yuqiuw:
Answers to your question:

- Yes this is only an issue where there are multiple inputs for a single label. Single text back is ok.
- This should be applied to all template packs but I didn't add it to all. Do you think I should do that? XD will be happy to.
- So this isn't template pack specific logic, again do you suggest me to apply it to all template packs (similar to this one)?

yuqiuw:
@smithdc1  Hi! Happy 2021! Any thoughts about this?

yuqiuw:
> @smithdc1 Hi! Happy 2021! Any thoughts about this?

Hi David, I'd like to push a little bit on this. Can I have one more review for this MR? Thank you for your attention!

smithdc1:
Hey @yuqiuw Thanks for your continued interest in this. I'm fairly unfamiliar with a11y and so on, so your help here is appreciated. I've loaded the [crispy-test-project](https://github.com/django-crispy-forms/crispy-test-project) with this patch and run the "lighthouse" tests from within Chrome. With this patch, I now see a long list of errors `aria` errors. :thinking: 

image

Secondly on your approach here you've made changes to the python code. We're trying to unpick the template packs from "core" the preferred solution if for template logic wherever possible. I think if we can get it working first, we can then think about this later.

yuqiuw:
Thank you for replying! I'll look into this!

yuqiuw:
@smithdc1  The original intention is to fix a11y problem only for multi-text input like this:
image

But it's also good to fix all a11y problems. Please check with the lighthouse that all aria-related problems are fixed. Also, the problems detected by Axe or Wave are mitigated intensely. Hope this helps, thank you!

yuqiuw:
image
Before this it was ~72

yuqiuw:
@smithdc1  Hi David, I'm still working on this and have enough bandwidth. Please help me review this and let me know if there're any places to improve/modify. XD thanks

smithdc1:
On mobile so sorry for typos / etc. 

Thanks for your continued effort here. 🎖️

I'm nervous about how big this change is, I don't think we need to add these attributes to every input. Bootstraps docs don't show this and it wasn't highlighted as part of the recent accessibility review of Django where there are a few tickets open, here are a couple of them. 

https://code.djangoproject.com/query?status=assigned&status=new&description=~Accessibility&desc=1&order=id

On the specific of split date time field I'm not even sure how this should work in standard Django. I asked on the forum but didn't get a response. Do you have any thoughts here?

https://forum.djangoproject.com/t/accessibility-for-splitdatetimefield/6871

There is the excellent crispy-forms-gds package. I suspect there'll be something in here that could help with this issue, but needs some time to investigate. 🕵️ 

https://github.com/wildfish/crispy-forms-gds

Finally I wonder if part of this is just layout of multiwidgets generally. A number of issues here, for example. I've looked at this a number of times and haven't yet worked out the right solution. I suspect the GDS package probably has an answer for this as well...

https://github.com/django-crispy-forms/django-crispy-forms/issues/831

yuqiuw:
@smithdc1  Thank you for your timely response! Really appreciate that.

For the first worry that "I don't think we need to add these attributes to every input." But this is necessary to fix all Aria errors shown in lighthouse.  Maybe it's a better idea to wait for the Django project to fix it... but since there have not been related discussions/issues posted there, I think it's fine for us to add it in crispy forms to provide better accessibility.

The split date-time field will just show up as your code is in your thread https://forum.djangoproject.com/t/accessibility-for-splitdatetimefield/6871 (and I appreciate that you brought this question up in the forum) My thought is the label is meant to be for both of these fields, so using aria-labelledby will solve this nicely (and thus my code is surrounding that)

And I agree that this could be just layout of multiwidgets generally, since it's because "for" can only point from one label to one input. Again I think "aria-labelledby" could solve this by one to multi mappings.

Briefly looked into the gds repo, it seems that they used "aria-descirbedby" to solve this. From my point of view that's kind of the same solution as mine?
image

Finally, I think given that this pull request is an improvement anyway, I still want to push it to merge XD. Thank you again for following up!

smithdc1:
So it looks like gds is using `described-by` for help text, that's great. I think we should do the same and is recommended by the bootstrap docs. 

I am not sure that every input needs an `aria-labelledby`, for example with your patch we get the following. However, this input will pass the accessibility tests without the aria label as we have declared it [explicitly](https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-explicitly) via `for` and `id` on the `label` and `input`. :thinking: 

``` html

<div id="div_id_default_form-text_input" class="form-group"> 
    <label for="id_default_form-text_input" class=" requiredField">
        Text input<span class="asteriskField">*</span> 
    </label>
    <div class=""> 
        <input type="text" name="default_form-text_input" class="form-control-lg textinput textInput form-control" 
         aria-labelledby="id_default_form-text_input_label" required="" id="id_default_form-text_input">
     </div>
</div>
```

yuqiuw:
@smithdc1  Oh sorry I missed your comment! You're right that not every input needs an `aria-labelledby`, only the input with sibling(s) needs it since all these input will point to the same label. But I think it's fine to have label and input(s) to point to each other

smithdc1:
crispy-forms 2.x will drop all the template packs in favour of stand alone template packs.  See https://github.com/django-crispy-forms/

I'd be more than happy to review accessibility PRs to those repos.

PR #1098, opened 2020-12-16
Document how to build documentation

cpina:


docs/contributing.rst
```diff
@@ -30,6 +30,18 @@ Django-crispy-forms comes with git hook scripts. These can be installed by runni
...
 
+Build the documentation locally
+===============================
+
+If you make documentation changes they can be seen locally.
+In the directory where you cloned ``django-crispy-forms``::
+
+    cd docs
+    pip install -r requirements.txt
+    make html
```
smithdc1, line 41: Or `make.bat html` on windows. It took me a good few weeks to figure out how to install `make` on windows when I was first learning. Could just add as a separate line with a comment for the os is applies to? 

See here for reference. 

https://docs.djangoproject.com/en/dev/internals/contributing/writing-documentation/

cpina: Actually, thanks to `make.bat` existing in the `docs`: I think that the current commands would also work on Windows. I guess that Django specifies for clarity though. I will check on a VirtualBox out of curiosity.

I'm happy to add a line to avoid confusion :-)

smithdc1: If it works then I'm not sure we need to be that explicit?

cpina: Confirmed: on Windows it works following the instructions as they are now. Perhaps as it is now is good to go then?

smithdc1: I agree.

smithdc1 (approved):

PR #1099, opened 2020-12-16
Fix `data-parent` in the AccordionGroup to refer to the Accordion id

cpina:
The Accordion unique ID is generated on the constructor of
Accordion and passed to its AccordionGroup.

### DIFF
--- a/crispy_forms/bootstrap.py
+++ b/crispy_forms/bootstrap.py
@@ -342,13 +342,19 @@ class Accordion(ContainerHolder):
 
     template = "%s/accordion.html"
 
-    def render(self, form, form_style, context, template_pack=TEMPLATE_PACK, **kwargs):
-        content = ""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
 
-        # accordion group needs the parent div id to set `data-parent` (I don't
-        # know why). This needs to be a unique id
+        # Accordion needs to have a unique id
         if not self.css_id:
             self.css_id = "-".join(["accordion", str(randint(1000, 9999))])
+
+        # AccordionGroup need to have 'data-parent="#Accordion.id"'
+        for accordion_group in args:
+            accordion_group.data_parent = self.css_id
+
+    def render(self, form, form_style, context, template_pack=TEMPLATE_PACK, **kwargs):
+        content = ""
 
         # Open the group that should be open.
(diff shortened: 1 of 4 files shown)


crispy_forms/bootstrap.py
```diff
@@ -342,14 +342,20 @@ class Accordion(ContainerHolder):
...
 
-        # accordion group needs the parent div id to set `data-parent` (I don't
-        # know why). This needs to be a unique id
+        # Accordion needs to have a unique id
         if not self.css_id:
             self.css_id = "-".join(["accordion", str(randint(1000, 9999))])
 
+        # AccordionGroup need to have 'data-parent="#Accordion.id"'
+        for accordion_group in args:
+            accordion_group.data_parent = self.css_id
```
cpina, line 354: In this line is where we could check that `accordion_group` doesn't have an existing `data_parent` and warn if it's the case as they should not be re-used. Is this worth it? How to warn?

At least we should add one line in the documentation saying to not re-use `AccordionGroup`. The warning I'm not sure. Is there any other case in `django-crispy-forms` with a warning? I can't remember now :-)

cpina:
On Friday I'll add an `Accordion` in `crispy-test-project` for easier testing. I have it locally but I need to do a bit of clean up.

crispy_forms/bootstrap.py
cpina: I've had a look at what happens if an `AccordionGroup` is re-used.

Because `AccordionGroup`s would have fields and they would be re-used `django-crispy-forms` shows a logging message (or an exception depending on `CRISPY_FAIL_SILENTLY`): `WARNING:root:A field should only be rendered once: text_input_accordion5`

If we wanted we could warn the user to avoid re-using `AccordionGroup` but probably they will not do or see the error avoid re-using fields. So I think that it might not be necessary? No strong opinion though.

cpina:
> On Friday I'll add an `Accordion` in `crispy-test-project` for easier testing. I have it locally but I need to do a bit of clean up.

This was done on this PR: https://github.com/django-crispy-forms/crispy-test-project/pull/23/

smithdc1:
Hi @cpina thanks for working on this. This looks great, I like this solution. :+1: 

I read your comment about the error messages. I don't think it is too much of a concern so let's have this as it is.

Issue #663, opened 2016-12-06
Override form_class for specific field


bluesurfer:
From the docs:

> The way you do horizontal forms in Bootstrap version 3 is setting some col-lg-X classes in labels and divs wrapping fields. This would mean a lot of hassle updating your layout objects for settings these classes, too much verbosity. Instead some FormHelper attributes have been added to help you easily achieve this. You will need to set only three attributes:
```helper.form_class = 'form-horizontal'
helper.label_class = 'col-lg-2'
helper.field_class = 'col-lg-8'
helper.layout = Layout(
    'email',
    'password',
    'remember_me',
    StrictButton('Sign in', css_class='btn-default'),
)
```

What if I want to use a different grid layout for a specific field? Let's say that I have a standard address form and want to wrap more than one one field with a different grid?

Something like this:
![image](https://cloud.githubusercontent.com/assets/4722772/20920637/03facdc6-bba1-11e6-8eba-12ef258e43bc.png)

Taken from here http://bootsnipp.com/snippets/featured/address-details-modal-form

tomedelliott:
You should be able to do something like:
```python
helper.label_class = 'col-lg-2'
helper.field_class = 'col-lg-8'
helper.layout = Layout(
    'line1',
    'line2',
    'city',
    Field('state', css_class="col-lg-2"),
    Field('postcode', css_class="col-lg-2"),
    country,
    StrictButton('Sign in', css_class='btn-default'),
)
```

bluesurfer:
That would just change the field's CSS class and it does not affect the row (fields are not inline)

doulix:
I think I found an acceptable solution:

```
self.helper.layout = Layout(
    Div(
        Div('inlineField1',css_class='col-md-6',),
        Div('inlineField2',css_class='col-md-6',),
        css_class='row',
    ),
    'other_fields',
    ...,

    FormActions(
        Submit('submit', 'Submit'),
    ),
)```

PR #1101, opened 2020-12-21
Allow label_class and field_class on a Field

Fasand:
... which overrides the one set on FormHelper within the template.
**Relevant issues**: #663, #348

When a `field_class` or `label_class` is set on a FormHelper, it is automatically applied to every field within the Layout, which makes variable grid layouts a nightmare.
Being able to specify each of these attributes globally on a FormHelper as the default and then on specific fields solves this issue where, e.g. most fields are laid out in a grid and one field stands out on top and spans the full width.

The solution uses a similar approach to how a `wrapper_class` is specified on a Field, and it essentially overwrites whatever was set by the FormHelper within the template.

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -440,6 +440,8 @@ class Field(LayoutObject):
             else:
                 self.attrs["class"] = kwargs.pop("css_class")
 
+        self.label_class = kwargs.pop("label_class", None)
+        self.field_class = kwargs.pop("field_class", None)
         self.wrapper_class = kwargs.pop("wrapper_class", None)
         self.template = kwargs.pop("template", self.template)
 
@@ -451,6 +453,11 @@ class Field(LayoutObject):
             extra_context = {}
         if hasattr(self, "wrapper_class"):
             extra_context["wrapper_class"] = self.wrapper_class
+        # Overwrite form-specified label and field class with field-specific
+        if hasattr(self, "label_class"):
+            extra_context["label_class"] = self.field_class
+        if hasattr(self, "field_class"):
+            extra_context["field_class"] = self.field_class
 
         template = self.get_template_name(template_pack)
 


crispy_forms/layout.py
```diff
@@ -451,6 +453,11 @@ def render(self, form, form_style, context, template_pack=TEMPLATE_PACK, extra_c
             extra_context = {}
         if hasattr(self, "wrapper_class"):
             extra_context["wrapper_class"] = self.wrapper_class
+        # Overwrite form-specified label and field class with field-specific
+        if hasattr(self, "label_class"):
+            extra_context["label_class"] = self.field_class
```
cpina, line 458: I haven't looked into detail but should `self.field_class` be `label_class` here? If not let me know and I'll come back to this properly.

cpina (commented):
Thanks @Fasand for looking into this!

I'll admit that I haven't set `label_class` / `field_class` in the helper method myself.

I think that the approach in the PR makes sense (I've left a comment in the ticket). I'll try to test it.

In any case, before merging it we would need to have unit tests, have you had a look at them?

@smithdc1: have you looked into the related tickets before?

smithdc1:
Hey @Fasand -- Thanks for this. Sounds like a reasonable proposal to me and I can see this being helpful to other template packs that may require more flexibility.  

> In any case, before merging it we would need to have unit tests, have you had a look at them?

Yes, please can we have some tests to go along with this. This will help me to sit down with this properly.

philippemilink:
Hello,

This PR brings the feature I was looking for ! 

However, to make it work, I had to add a test to check if `self.label_class` and `self.field_class` are not null:
```python
# Overwrite form-specified label and field class with field-specific
if hasattr(self, "label_class") and self.label_class is not None:
    extra_context["label_class"] = self.label_class
if hasattr(self, "field_class") and self.field_class is not None:
    extra_context["field_class"] = self.field_class
```

It would be great if this feature could be merged and released !

smithdc1:
Hey @philippemilink 

This needs tests and docs. I must admit I'm still not entirely sure what problem this solves! Hopefully tests and docs will help me understand this better. 

Do you fancy picking this change up?

smithdc1:
Closing to inactivity. Happy to take a look at this if someone wishes to pick up the PR and address the comments above.

Issue #1097, opened 2020-12-15
Update copyright year


cpina:
At the bottom of the documentation: https://django-crispy-forms.readthedocs.io/en/latest/

It says `© Copyright 2009-2013, Miguel Araujo and Daniel Greenfeld Revision 121419fe.`

When I'm screening packages and I see very old copyright years my alarms go off: is this package not updated for 7 years? Is it not well maintained? Are they not doing releases? Not maintaining the docs?

Should we just change the copyright year? (https://github.com/django-crispy-forms/django-crispy-forms/blob/master/docs/conf.py#L48). Or do like Django, something like: `© Copyright 2009-2020, Miguel Araujo, Daniel Greenfeld and individual contributors Revision xxxxx.`

smithdc1:
Yes, let's update it. :+1: 

- I think adding and contirutors is nice (not sure about individual, but I'm not too sure what the norm is here). 
- Let's corrent Daniel's name. 
- We should also update the licence?  https://github.com/django-crispy-forms/django-crispy-forms/blob/master/LICENSE.txt#L1

cpina:
I had a quick look:
 * https://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/#putting-it-all-together : would suggest `© Copyright 2009-2020, Miguel Araujo, Daniel Greenfeld and contributors`
 * `individual`: not needed for the reading that I've done this morning. When I checked the Django one (https://docs.djangoproject.com/en/3.1/) probably `individual` is there as opposed as the foundation
 *  Yes for changing the copyright in the license as well

If others are happy I can create a PR next days.

smithdc1:
Fixed in fcbddb416d31b6528bb12f9ad981a088442e8b25 :tada:

PR #1106, opened 2021-01-09
Update copyright notices

cpina:
See issue #1097

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,7 +45,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = 'django-crispy-forms'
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -45,7 +45,7 @@
 
 # General information about the project.
 project = 'django-crispy-forms'
-copyright = '2009-2013, Miguel Araujo and Daniel Greenfeld'
+copyright = '2009-2021, Miguel Araujo, Daniel Feldroy and contributors'
```
cpina, line 48: Since it's Python code we could have `datetime.now().year` . I've left it to be changed manually for now (so we change the LICENSE as well, etc.).

The LICENSE already had "contributors" but not Daniel - all seems good now.

PR #1114, opened 2021-02-13
Fixed input group size in Bootstrap 4

smithdc1:
@cpina this is still WIP, but thought it worth sharing more publicly my progress with this. Notes:

- I think your solution for the button groups by adding an extra argument of `input_size` is spot on. I think we need this for the append inputs as well
- This is due to bs3 and 4 having different behavior. On bs3 we want the spans and input to have a class, for bs4 we want the wrapping div to have the class (and only the wrapping div). 
- Adding this additional argument allows us to manage this flexibility -- I think we'll also need this for the bs5 pack, so will make things easier there.
- `Field_with_buttons` I've not really looked at. Tempted to chop it from this PR and get the rest of it merged.

### DIFF
--- a/docs/layouts.rst
+++ b/docs/layouts.rst
@@ -205,7 +205,7 @@ These ones live under module ``crispy_forms.bootstrap``.
 .. image:: images/form_actions.png
    :align: center
 
-- **AppendedText**: It renders a bootstrap appended text input. The first parameter is the name of the field to add appended text to, then the appended text which can be HTML like. There is an optional parameter ``active``, by default set to ``False``, that you can set to a boolean to render appended text active::
+- **AppendedText**: It renders a bootstrap appended text input. The first parameter is the name of the field to add appended text to, then the appended text which can be HTML like. There is an optional parameter ``active``, by default set to ``False``, that you can set to a boolean to render appended text active. Accepts the `input_size`_ argument::
 
     AppendedText('field_name', 'appended text to show')
     AppendedText('field_name', '$', active=True)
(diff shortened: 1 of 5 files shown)


smithdc1:
OK -- I've trimmed this down to just the append and related classes. We can come back to "with buttons" as a separate change.

cpina:
I'll review it, latest on the next weekend and probably and earlier :-)

smithdc1:
No rush! 🙂

docs/layouts.rst
```diff
@@ -257,7 +257,7 @@ These ones live under module ``crispy_forms.bootstrap``.
 .. image:: images/field_with_buttons.png
    :align: center
 
-- **Tab & TabHolder**: ``Tab`` renders a tab, different tabs need to be wrapped in a ``TabHolder`` for automatic JavasSript functioning, also you will need ``bootstrap-tab.js`` included in your static files::
+- **Tab & TabHolder**: ``Tab`` renders a tab, different tabs need to be wrapped in a ``TabHolder`` for automatic javascript functioning, also you will need ``bootstrap-tab.js`` included in your static files::
```
cpina, line 260: Mmm... Django documentation uses `JavaScript` - I wonder if the change is on purpose or mistake? I think that in django-crispy-forms we tend to have JavaScript as well.

docs/layouts.rst
```diff
@@ -301,6 +301,21 @@ These ones live under module ``crispy_forms.bootstrap``.
...
+Input group size
+----------------
+
+**Input group size**: Adjust the size of an input group (``AppendedText``, ``PrependedText``, ``PrependedAppendedText``) with the appropriate CSS class::
+
+    # Bootstrap 3 - Inputs and spans need size class. Use `css_class`.
+    PrependedText('field_name', StrictButton("Go!"), css_class="input-sm")
+    PrependedText('field_name', StrictButton("Go!"), css_class="input-lg")
+
+    # Bootstrap 4 - Wrapping div needs size class. Use `input_size`.
```
cpina, line 315: If we don't pass `input_size` what happens? I thought that it all still works with the default Bootstrap size, right?

Perhaps this could be added here stating that `input_size` is optional, but that if it's passed then it's used in certain way?

cpina (commented):
I've left two small comments about documentation - all the code looks good to me!

docs/layouts.rst
smithdc1: @cpina -- you are right but I'm struggling to find the right way to say this succinctly. Can you have a look at my next itteration here?

cpina (approved):

docs/layouts.rst
cpina: I like it - thanks!

Issue #1111, opened 2021-02-10
custom-select class disapear after wrapping in PrependedText


iGoosiatin:
* Package version: 1.11.0
* Django version: 3.0.5
* Python version: 3.9.0
* Template pack: bootstrap4

### Description:

Generating custom layout, both 'owner' and 'currency' are selects
```
class AccountForm(forms.ModelForm):
    def __init__(self, *args, **kwargs):
        super(AccountForm, self).__init__(*args, **kwargs)
        self.helper = FormHelper(self)
        self.helper.form_show_labels = False
        self.helper.field_class = 'input-group-lg'

        # owner = PrependedText('owner', 'Owner:')
        currency = PrependedText('currency', 'Currency:')

        submit = Submit('submit', 'Submit')
        submit.field_classes = "btn btn-outline-dark btn-block btn-lg"

        self.helper.layout = Layout(

            'owner',
            currency,
            submit
        )

    class Meta:
        model = Account
        fields = ['owner', 'currency']
```
Generated HTML - no "custom-select" class for 'currency'
```
<select name="owner" class="select form-control custom-select" id="id_owner"> 
        <option value="3" selected="">Name3</option>
       <option value="2">Name2</option>
</select>
<select name="currency" class="select form-control" id="id_currency">
   <option value="€" selected="">EUR</option>
   <option value="$">USD</option>
</select>
```

smithdc1:
Thanks 👍this is related to #1093. 

Let's keep this as I think this is a smaller part of the issue and we may be able to solve this bit a little easier. 🤷‍♀️

cpina:
I should have time over the weekend to look at this :-) if anyone goes ahead before mention it here...

cpina:
I'm thinking of a better solution (I'm thinking of a way to refactor the 4 current `{% crispy_field ... %}` calls in `field.html` but this works: https://github.com/cpina/django-crispy-forms/commit/6b2df6c1e1b295d8bccb5213e895de30f1c48c1b

Refactoring two of them could be done adding a new template and including it. I'll think a bit more.

smithdc1:
@cpina the commit you linked to is exactly what I was thinking.

cpina:
> @cpina the commit you linked to is exactly what I was thinking.

[...]

PR #1116, opened 2021-02-20
Fixed custom-select missing in select fields that have use_custom_con…

cpina:
…trol and prepended/appended text ((#1111)

closing #1111


smithdc1 (approved):
LGTM. Thank you!

PR #1128, opened 2021-03-14
Added html parsing tools for testing

smithdc1:
This allows the expected html to be stored in an HTML file. Tests can then assert the layout of the whole form.

I've been using this method with the bootstrap5 template pack and have been enjoying working in this way. 
I'd like to introduce this here as I think it is helpful generally, I find that tests are more readable as you can see all of the output. It's also then easier to have a discussion about if there is an issue as you can point to the line in the html code that is in question. 

More specifically I think it will be helpful to prove the optgroup filter and upcoming changes are ok as we'll clearly be able to see the changes that happen (hopefully none :crossed_fingers: ) which I think could be missed with the current approach.

### DIFF
--- a/crispy_forms/tests/results/utils_test.html
+++ b/crispy_forms/tests/results/utils_test.html
@@ -1,0 +1,7 @@
+<form method="post">
+    <div class="form-group">
+        <div class="custom-checkbox custom-control" id="div_id_is_company"><input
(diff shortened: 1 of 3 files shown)


carltongibson (approved):
Hey @smithdc1 — I think this is a good move. Having the HTML as strings in the tests has been a pain point ≈forever — so attacking that has to be a win. 👍

smithdc1:
Thanks Carlton!

PR #1130, opened 2021-03-15
Fixed #1129: render all widget attributes for file fields in bootsrap4

### LINKED ISSUE #1129: bootstrap 4 field_file template does not render custom classes and attributes from the widget
* Package version: 1.11.1
* Django version: 3.1.7
* Python version: 3.9.2
* Template pack: bootstrap 4

### Description:

The bootstrap 4 `field_file.html` template does not respect custom widget classes and custom attributes coming from the widget.

E.g. consider the following code:
```
class MyCustomClearableFileInput(forms.ClearableFileInput):
    def build_attrs(self, *args, **kwargs):
        attrs = super().build_attrs(*args, **kwargs)

        attrs["data-foo"] = "bar"
        return attrs


class myForm(forms.Form):
    files = forms.ImageField(
        label=_("images"),
        widget=MyCustomClearableFileInput(
            attrs={"accept": "image/png", "multiple": True, "class": "my-custom-class"}
        )
    )
```
The html output will look like this:
```
<input type="file" name="files" class="custom-file-input" accept="image/png" multiple="">
```
But what I would expect is:
```
<input type="file" name="files" class="custom-file-input my-custom-class" accept="image/png" multiple="" data-foo="bar">
```

I have worked around the problem by overwriting and adapting the template to use `widget.data.attrs.items` instead of `field.field.widget.attrs.items` and to also respect `widget.data.attrs.class` if it's there. 
If you can confirm that the current behavior is not intended like this, then I'm happy to submit a PR with my changes if you like.

Btw, in case you are wondering about use cases: I found the issue because I'm using [django-s3file](https://github.com/codingjoe/django-s3file) which adds a lot of custom attributes by [sneaking a mixin](https://github.com/codingjoe/django-s3file/blob/master/s3file/apps.py) into the `ClearableFileInput` widget and without my workaround it won't render those attributes.

### Preferably also include:

- [x] Example Django Crispy Forms code
- [ ] Screenshots
- [x] Actual HTML generated
- [x] Expected HTML

anneFly:
I've explained the issue a bit more in #1129 .
The main change is that the template now uses `widget.data.attrs` and that `widget.data.attrs.class` is explicitly rendered when it's there. Please let me know if you need more/different test coverage or different naming.

### DIFF
--- a/crispy_forms/templates/bootstrap4/layout/field_file.html
+++ b/crispy_forms/templates/bootstrap4/layout/field_file.html
@@ -27,7 +27,7 @@
     </div>
 {% endif %}
     <div class="form-control custom-file{% if field.errors %} is-invalid{%endif%}" style="border:0">
-        <input type="{{ widget.data.type }}" name="{{ widget.data.name }}" class="custom-file-input{% if field.errors %} is-invalid{%endif%}" {% if field.field.disabled %}disabled{% endif %} {% for name, value in field.field.widget.attrs.items %}{% if value is not False %} {{ name }}{% if value is not True %}="{{ value|stringformat:'s' }}"{% endif %}{% endif %}{% endfor %}>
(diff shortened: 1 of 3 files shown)


smithdc1:
Hi Anne -- Thanks for this, it looks good to me. 

The tests don't seem to be running, I've tried to restart them but it seems a bit stuck :detective:  -- https://github.com/django-crispy-forms/django-crispy-forms/actions/runs/655500429 

If it's still not completed in the morning, could you try rebasing to see if that spurs it into life?

anneFly:
@smithdc1 looks like the tests have been running by now. All green :)

smithdc1:
@anneFly that's good news. Thanks again for the contribution!

PR #1135, opened 2021-04-07
Confirmed support for Django 3.2

smithdc1:


carltongibson (approved):
Super! 🎁


Fancy dropping Python 3.5? Folks aren't really using it **and** updating.

smithdc1:
I was thinking about Python 3.5 as well. I'll open a separate PR to stop testing against Python 3.5.

PR #1137, opened 2021-04-08
Improved Optgroups tests

smithdc1:
This is part of the work to use the new `optgroups` filter. We need better test coverage of the rendered HTML so we can easily see any changes when I change the templates. 

I've also structured the existing tests better in a separate commit. The logic is now each batch of HTML "result" files are saved under the convention of `{template pack}/{test file name}/{test name}{test identified (if required)}.html`.

Finally, it seems we have good coverage of checkboxes, but we should have the same for radios.


smithdc1:
Let's get #1141 in first.

smithdc1:
Awsome, all green ticks. I'm pleased that writing tests is getting easier and more robust.

PR #1141, opened 2021-04-09
Zero index checkbox and radio in bootstrap templates

smithdc1:
When we move to the new optgroups filter the ids for the inputs will be start from zero rather than one. I'm making this change now so we have a smaller diff later on. (or atleast that's the intention!)

I'm wondering if this is backwards incompattible change? Any comments?


smithdc1:
I have been thinking about this for the past few days. 

- DCF is currently doing something different to Django itself, we're therefore reverting to Django's own style. 
- I've not seen any folk have an issue with there being a difference today, this makes me think folk are not impacted by this. 

On that basis, I think we should continue.

PR #1149, opened 2021-05-01
Fixed update_attributes() to use same HTML5 attributes behavior for "_" and "-"

SantaFox:


smithdc1:
Hey, are you still interested in this change. I'm concerned that the tests fail and that this will cause backwards compatibility issues for folk. We've seen recently that folk rely on the current behaviour even where the usage is not as intended.

PR #1158, opened 2021-05-11
Added guidance for custom widgets

### LINKED ISSUE #768: CKEditorWidget incompatible with crispy forms
* Package version: 1.7.0
* Django version: 1.8
* Python version: 2.7

### Description:
There is an incompatibility with CKEditorWidget.
When using crispy form tabs with fields that have CKEditorWidget widget, the widget no longer works properly. It looks like a textarea withing a box so the rendering is completely broken.
Note: I am using bootstrap3.

### Preferably also include:

- [ ] Example Django Crispy Forms code
- [ ] Screenshots
- [ ] Actual HTML generated
- [ ] Expected HTML

### LINKED ISSUE #971: simple-captcha validation error is not shown
Hello,

When using a CaptchaField from simple-captcha module with crispy-forms, if I leave the field empty, or provide an invalid response, no error is displayed.
If I disable crispy-forms, it works.

I looked into the generated HTML and found the entry:
```html
<span id="error_1_id_captcha" class="invalid-feedback"><strong>Invalid response</strong</span>
```

The problem is, this CSS class has a "display: none" prop... So I overloaded my global CSS with the following:

```css
/* Workaround a bug in crispy-forms not showing validation error from simple-captcha */
.invalid-feedback {
    display: block;
}
```

And the validation error is now displayed successfully.

Best regards, Adam.

smithdc1:
We see a number of issues where folk are trying to use custom widgets in their form (#768, #971 etc). Given the myriad of things these could be I'm not sure there's a simple generic answer that could work. 

I think the best we can do is to guide folk to use a custom template, as this is feeling quite bespoke at this point? 

Closes #768 
Closes #971

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -23,6 +23,11 @@ Of course you can display form fields within columns, the same way you would do 
 .. _`example`: http://stackoverflow.com/questions/12144475/displaying-multiple-rows-and-columns-in-django-crispy-forms
 
 
+Using a custom widget
+~~~~~~~~~~~~~~~~~~~~~
+
+If you are using a custom widget you may find that django-crispy-forms does not render it correctly. In this case please use a custom template for that field, see :ref:`override templates`.
+
 .. _faq-general:
 
 General


carltongibson (approved):
Yep. 😃

A widget is at least 50% the actual HTML it generates. I think this is fair

smithdc1:
Thank you for the review! 👍

PR #1159, opened 2021-05-12
Fixed Bootstrap4 Field with buttons input size

smithdc1:
@cpina -- Could you have a look at this at some point for me? The first commit is just resturcuting the tests, the second commit is the real change and is much smaller. 

The `FieldWithButtons` class is a little different to the prepend / append inputs so I think the solution needs to be different as well. 

I had a look through the git history and can't quite work out the history of the `input_size` variable that's in the template at the moment. I'm also unsure how you would even use this? 

Key questions that I'm slightly unsure on are both related to way the `__init__` function is structured:

- Does this break backwards compatibility give I've introduced a new `__init__` function with a different structure. I _think_ by adding the new item as a key word only variable this is ok? 
- Is adding this item as a key word argument to `FieldWithButtons` the way folk would want to use it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Added support for Python 3.10
 * Dropped support for Django 3.1
 * Dropped support for Python 3.6
+* Added `input_size` argument to `FieldWithButtons` to allow customisation of the size of the input in the Bootstrap 4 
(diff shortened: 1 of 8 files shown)


cpina:
I'll have a look :-)

PR #1161, opened 2021-05-15
Updated uni-form link

### LINKED ISSUE #1050: link to uni_form template pack is probably wrong
The link to the uni_form template pack[1] in `docs/install.rst` redirects to the `development` branch of `uni_form`.  It seems like `django-crispy-forms` expects a CSS class called `uniForm`, which doesn't exist in this branch.

However, the `uniForm` class *does* exist in the https://github.com/draganbabic/uni-form/tree/uni-form-v-1-5 branch, which claims to be the one used in "most integrations".  I suspect `django-crispy-forms` expects this version, in which case, a simple documentation fix is all that's wanted.

[1] http://sprawsm.com/uni-form

smithdc1:
Fixes #1050

Given Uni-Form hasn't had a commit since 2013 it's looking a bit unmaintained to me. I wonder how much longer we should continue to package those templates? I guess putting it into it's own template pack would be the answer...

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -70,4 +70,4 @@ crispy-forms does not include static files. You will need to include the proper 
  
 
 .. _Django: https://djangoproject.com
-.. _`Uni-form`: https://sprawsm.com/uni-form
+.. _`Uni-form`: https://github.com/draganbabic/uni-form/tree/uni-form-v-1-5


smithdc1:
Hmm... tests failed after merging this. Looks like it is due to codecov failing 🤷. Will re-base another PR and see what happens.

PR #1162, opened 2021-05-21
Fixed compatibility with bootstrap5 template

lauri-openscop:
Whitout this offset in bootstrap 5 template render in "col-*-offset-*" and this doesn't fit with Bootstrap V5

### DIFF
--- a/crispy_forms/helper.py
+++ b/crispy_forms/helper.py
@@ -335,7 +335,7 @@ class FormHelper(DynamicLayoutHandler):
             "use_custom_control": self.use_custom_control,
         }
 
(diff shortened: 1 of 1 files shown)


smithdc1:
Thanks for the pull request. 

We're aiming to spit the template packs up so they are not dependent upon the core application. 

I therefore think we should look to find a way of making this work within the crispy-bootstrap5 package. Can we do replicate what's going on here in template logic? That would be the first place I'd have a look. 🕵️

smithdc1:
I see you've opened a PR on the bootstrap5 repo, that's great. I'll therefore close this one. ﻿

PR #1165, opened 2021-05-25
Version 1.12

smithdc1:
There is rather alot of code [on the shelf](https://github.com/django-crispy-forms/django-crispy-forms/milestone/16?closed=1) with 20 changes since the last release. I'm really pleased with what's in this one, resolving a long standing grouped inputs bug and confirming support for Django 3.2 are the benefits from a user perspective. 

The addition of the `optgroups` filter is a nice win from a maintenance perspective as it should simplify the templates somewhat.

- [x] Update the release date.


smithdc1:
I've created a regression in https://github.com/django-crispy-forms/django-crispy-forms/commit/a7c0c1501629a006d3dd9a79fa9eb117383e0c52

While checking the comment there that radios are completely broken I went back and did some more testing. While I can't yet replicate the users feedback (works for me) we're now getting a rather odd layout for error messages. 🕵️ 

image

smithdc1:
> I've created a regression in [a7c0c15](https://github.com/django-crispy-forms/django-crispy-forms/commit/a7c0c1501629a006d3dd9a79fa9eb117383e0c52)

Fixed in a5cef6878e78c6110584e69ee36f060e79d44828

PR #1167, opened 2021-06-03
Fixed error messages for checkboxes and radios

smithdc1:
This patch fixes the regression I caused when fixing grouped inputs. The error message was being shown below each option. I've also provided an image of what we have now vs the current release to show the previous behaviour being restored. 

This patch
image
image

v1.11.2
image
(grouped input's don't work here -- that patch is in the next release)


smithdc1:
Yay, the tests even passed on Django 2.2! 😄

Issue #1169, opened 2021-06-08
render_field function doesn't work with formset


wail-asad:
* Package version:1.11.0
* Django version:3.2.0
* Python version:3.7.6

### Description:

I used Django render_field method to return a form design of formset.
I know that there a method to get tabular design of form set by using `self.helper.template='crispy-tabuler-page'` but this method gives all fields of formsets  and ignore all design in `self.helper.layout`
 so I used render_field method to get html design of formset in the view.
 the problem is 
`Django Version:	3.2
Exception Type:	TypeError
Exception Value:	
list indices must be integers or slices, not str
Exception Location:	C:\0ERPFlex\ERPFlex_venv\lib\site-packages\django\forms\formsets.py, line 96, in __getitem__
Python Executable:	C:\0ERPFlex\ERPFlex_venv\Scripts\python.exe
Python Version:	3.7.6`

I make this fixed in the method **render_fields** :
  `if hasattr(form,'forms'):
                form=form.forms`
 
and the problem was solved but I want a this fixed in the next version in **render_fields**

smithdc1:
I'm not sure I understand your issue.

Happy to review a Pull Request with tests showing the behaviour you desire would be welcomed. 👍

wail-asad:
@smithdc1 could you open a branch with these issue to make pull request  I want also to make Layout for formset

smithdc1:
Please send the pull request to the main branch of this repository.

wail-asad:
@smithdc1 let me know if I can make push for a tabular formset design with js file for its actions.

PR #1173, opened 2021-06-11
Fixed render_field for formset (#1169)

wail-asad:


wail-asad:
@smithdc1 the coverage decrease is due to the test, if I make the test, Is will be executed by **codecov** before the merge? 
This doesn't encourage me to push the layout design for tabular formset.

PR #1182, opened 2021-08-15
Fix css_id for non-ASCII characters - allow unicode in slugify function

izabelakowal:
I've been working on a project that is translated into multiple languages. The bootstrap accordion groups haven't been functioning properly while the container name was translated to Japanese. I realized that the `css_id` of the container is slugified while allowing only ASCII characters.

### DIFF
--- a/crispy_forms/bootstrap.py
+++ b/crispy_forms/bootstrap.py
@@ -1,8 +1,8 @@
 from random import randint
 
 from django.template import Template
-from django.template.defaultfilters import slugify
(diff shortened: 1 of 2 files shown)


izabelakowal:
Fixed the isort issue.

smithdc1:
Thanks for the PR.

Do you think you could add a test for this?

izabelakowal:
@smithdc1 test case added.

smithdc1 (approved):
Thanks for the patch. 👍

PR #1192, opened 2021-10-30
Improved ButtonHolder init signature and doc string

smithdc1:
Most of the `__init__` signatures are `*args, **kwargs`, the method then pops them off as needed. 

I've been thinking what is best practice here as it makes it a little hard to understand what each class accepts. Here is a patch showing one of the more simple cases where I think writing out the kwargs in full will make the package easier to use as folk will get auto complete and so on? 

There's more complex cases, but thought best to start off with the "simple" scenario.

I'm interested in understanding more about conflicting design principles here. The giants who's shoulders we are standing on clearly wrote this for a very good reason! (albeit I wonder if some of this is differences over time, this was originally written about a decade ago!)

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -139,15 +139,45 @@ class ButtonHolder(LayoutObject):
     """
     Layout object. It wraps fields in a <div class="buttonHolder">
 
-    This is where you should put Layout objects that render to form buttons like Submit.
-    It should only hold `HTML` and `BaseInput` inherited objects.
+    This is where you should put Layout objects that render to form buttons
+    like Submit. It should only hold `HTML` and `BaseInput` inherited objects.
+
+    Attributes:
+        template : str
+            The default template which this Layout Object will be rendered
+            with.
+
+    Args:
+        fields : HTML or BaseInput
+            The layout objects to render within the ButtonHolder.
+
+        css_class : str, optional
+            Additional CSS classes to be included in the
+            button holder `<div>`.
+
+        css_id : str, optional
+            Sets a DOM id for the button holder. .
+
+        template : str, optional
+            Over rides the default template, if provided.
 
     Example::
 
+        # Within a form's `Layout`.
+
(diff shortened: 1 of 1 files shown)


smithdc1:
I've had a go at improving the doc string here. It now documents the arguments it accepts which I think is an improvement. Not sure if I've got this at the right level though.

crispy_forms/layout.py
```diff
@@ -139,24 +139,54 @@ class ButtonHolder(LayoutObject):
     """
     Layout object. It wraps fields in a <div class="buttonHolder">
 
-    This is where you should put Layout objects that render to form buttons like Submit.
-    It should only hold `HTML` and `BaseInput` inherited objects.
+    This is where you should put Layout objects that render to form buttons
+    like Submit. It should only hold `HTML` and `BaseInput` inherited objects.
+
+    Attributes:
+        template : str
```
smithdc1, line 146: Not sure if we should duplicate this as an attribute to the class and as an arg to __init__. Maybe just include in the `args?`.

crispy_forms/layout.py
```diff
@@ -139,24 +139,54 @@ class ButtonHolder(LayoutObject):
...
+    Args:
+        fields : HTML or BaseInput
+            The layout objects to render within the ButtonHolder.
+
+        css_class : str, optional
+            Additional CSS classes to be included in the
+            button holder `<div>`.
+
+        css_id : str, optional
+            Sets a DOM id for the button holder. .
```
smithdc1, line 159:
```suggestion
            Sets a DOM id for the button holder.
```

crispy_forms/layout.py
```diff
@@ -139,24 +139,54 @@ class ButtonHolder(LayoutObject):
...
+
+        template : str, optional
+            Over rides the default template, if provided.
```
smithdc1, line 162:
```suggestion
            Overrides the default template, if provided.
```

PR #1199, opened 2021-12-15
Honor field auto_id when rendering radio

nijel:
This is needed when using custom auto_id to render same field on the page multiple times.

### DIFF
--- a/crispy_forms/templates/bootstrap3/layout/radioselect.html
+++ b/crispy_forms/templates/bootstrap3/layout/radioselect.html
@@ -6,8 +6,8 @@
 
     {% for choice in field.field.choices %}
       {% if not inline_class %}<div class="radio">{% endif %}
-        <label for="id_{{ field.html_name }}_{{ forloop.counter0 }}" class="{% if inline_class %}radio-{{ inline_class }}{% endif %}">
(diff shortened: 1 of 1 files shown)


smithdc1:
Thanks for the patch. 👍 

I have a few questions / comments, appreciate thoughts. 

Would this change also apply to multiple checkboxes? 

While we approach these slightly differently in the bootstrap 4 template pack, I wonder if it could be a similar issue there? 

It would be useful to add a small test to ensure we don't regress on this in the future.

nijel:
Generally, I don't think it is a good idea to generate the element id in the template when it is already provisioned by the field itself. I haven't looked at other occurrences or template packs, though.

nijel:
Yes, checkboxselectmultiple should use this as well and other template packs also.

It would be good to agree on a solution here, as this is apparently problematic area touched several times (this includes myself five years ago):

* https://github.com/django-crispy-forms/django-crispy-forms/pull/758
* https://github.com/django-crispy-forms/django-crispy-forms/pull/783
* https://github.com/django-crispy-forms/django-crispy-forms/pull/892

For now, we're shipping patched template in Weblate to address this (https://github.com/WeblateOrg/weblate/commit/524dbeeec7b6b0156ac2cc0d128a8d28cffc5e71, patch for checkboxselectmultiple will land in https://github.com/WeblateOrg/weblate/pull/7156).

smithdc1:
I think the way forward here would be to restructure the template to use the `optgroups` filter as in the bootstrap4 template. I appreciate doing this is much more effort than the change you propose here. 

`crispy_forms/templates/bootstrap4/layout/radioselect.html`

The optgroups filter will give the same attributes as provided by Django core. 


Are you able to confirm the structure of the bs4 template gives you the structure you need?

nijel:
That indeed looks as cleaner approach. 

Using `crispy_forms/templates/bootstrap4/layout/radioselect.html` seems to work just fine - the `auto_id` passed to Django Form is correctly used for `input` and `label` tags.

smithdc1:
Superseded by https://github.com/django-crispy-forms/django-crispy-forms/pull/1262

PR #1210, opened 2022-01-23
Improved docstring and __init__ signatue of BaseInput and its subclasses

smithdc1:
Not for now, but I'm starting to think about a 2.0 release. One of the things that I'd like to do is remove as far as reasonable the **args** and **kwargs** signatures for something which lists out all of the arguments. 

While it is not the intent for these changes to backwards incompatible there's more risk that these changes could introduce a breaking change for someone. 

At the same time I'd also like to try and improve the doc strings. I've had a first draft, likely they could be improved, but maybe these are good enough. 

Too much to show all of it, but here is how vscode now adds it's help hints. 

image
image

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -170,20 +170,45 @@ class ButtonHolder(LayoutObject):
 class BaseInput(TemplateNameMixin):
     """
     A base class to reduce the amount of code in the Input classes.
+
+    Attributes
+    ----------
+    template: str
+        The default template which this Layout Object will be rendered
+        with.
+
+    Parameters
+    ----------
+    name : str
+        The name attribute of the button.
+    value : str
+        The value attribute of the button.
+    css_id : str, optional
+        A custom DOM id for the layout object. If not provided the name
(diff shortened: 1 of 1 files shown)


smithdc1:
@carltongibson -- As a general principle are you ok with me proceeding with this type of change? I'm removing improving the __init__ signature to avoid the slightly untransparent `*args, **kwargs` and aiming to significantly increase the doc strings. 

We'll include these as part of a 2.0 release. (I'm not expecting backwards compat issues here but who knows what folk are actually doing downstream)

carltongibson (commented):
Yes, I think this looks great. 

Under normal conditions nothing should break, the tests pass after all, but maybe folks subclassing hit some weirdness. 

That's OK, it's a major version. 

What I'd try to do though is begin the release notes as you go, rather than leaving them to the end, in order to write detailed thoughts, and any migration notes, whilst it's fresh in your mind. 

Make sense?

smithdc1:
Yes, that makes sense. Thank you for the guidance!!

PR #1211, opened 2022-01-23
Removed uni-form link and added tailwind in readme.

### LINKED ISSUE #1200: Link to uni-form is broken
In the resdme, then link to https://sprawsm.com/uni-form

Is broken

smithdc1:
Fixes #1200


smithdc1:
I guess the link is https://github.com/draganbabic/uni-form

But that's not had a commit since 2013 so I'm putting it in the unmaintained category. 

Should be something we could look to unpick from in a 2.0 release. 🤔

PR #1216, opened 2022-01-27
Enhanced __init__ signature of Button and Reset

smithdc1:
Follow up to ea802b82d94b2871a115ea9adf771704175ab3a4

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -344,9 +344,9 @@ class Button(BaseInput):
 
     input_type = "button"
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, name, value, *, css_id="", css_class=None, template=None, **kwargs):
         self.field_classes = "button" if get_template_pack() == "uni_form" else "btn"
-        super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


crispy_forms/layout.py
```diff
@@ -460,9 +460,9 @@ def __init__(self, *args, **kwargs):
 
     input_type = "reset"
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, name, value, *, css_id="", css_class=None, template=None, **kwargs):
```
smithdc1, line 463: Oh my. I've spent _hours_ looking at this and getting muddled. The Fieldset class (next class down) has `css_id=None, css_class=""` which is the opposite of what this set of classes does.

I think we can make `css_id=None` and generate the `id` in the init method. Currently if a custom `id` is not provided the `id` gets generated on each render (including a slugify call). Let's do that in the `__init__` and avoid that. Separare PR for that.

I think ideally we want to make the signatures consistant. Having different `None` vs `""` is confusing for me, let alone others!

As it happens I guess it doesn't matter too much as they both get treated the same in templates.

PR #1221, opened 2022-01-29
Removed Uni-Form Template Pack

smithdc1:


CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 # CHANGELOG for django-crispy-forms
 
 ## 2.0 (tbc)
+* Removed uni-form template pack. Uni-Form specific classes previously added to every template pack e.g. `textInput` and are now removed.
+  If you require these classes then the previous behaviour can be restored by adding the following to `CRISPY_CLASS_CONVERTERS` in your settings file.
+  ```
```
smithdc1, line 6: Can we use GitHub flavoured markdown?

crispy_forms/tests/test_form_helper.py
```diff
@@ -548,54 +530,6 @@ def test_helper_std_field_template_no_layout():
         assert html.count('id="div_id_%s"' % field) == 1
 
 
-@only_uni_form
-def test_form_show_errors():
```
smithdc1, line 552: These two error tests have bs3/bs4 equivalents.

crispy_forms/tests/test_layout.py
```diff
@@ -489,21 +490,12 @@ def test_layout_composition():
...
-    assert 'input type="submit"' in html
-    assert 'name="Save"' in html
-    assert 'id="custom-div"' in html
-    assert 'class="customdivs"' in html
-    assert "last_name" not in html
-
-
-@only_uni_form
-def test_second_layout_multifield_column_buttonholder_submit_div():
+    # Bootstrap 4 does not contain a multifield template
```
smithdc1, line 493: I think there is value in testing of this layout, so switched to bootstrap3. I'd run for bootstrap3 and 4 but this layout is currently invalid in bootstrap 4. But that's a separate issue/pr.

docs/layouts.rst
```diff
@@ -169,13 +165,7 @@ Renders to::
...
     )
 
-
-Uni-form layout objects
-~~~~~~~~~~~~~~~~~~~~~~~
-
-These ones live in module ``crispy_forms.layout``. Probably in the future they will be moved out to a ``uni_form`` module:
-
-- **ButtonHolder**: It wraps fields in a ``<div class=”buttonHolder”>``, which uni-form positions in a nice way. This is where form's submit buttons go in uni-form::
+- **ButtonHolder**: It wraps fields in a ``<div class=”buttonHolder”>``, this is a legacy layout object from the ``uni-form`` template pack::
```
smithdc1, line 168: The docs recommend using `FormActons` but these templates exist in the bootstrap template packs. Therefore folk will be using them and therefore think it's ok to keep these even if we acknowledge they are not the recommended approach. I don't think there is a need to introduce a potential breaking change at this stage for this.

CHANGELOG.md
smithdc1: It looks highlighted correctly.

PR #1246, opened 2022-03-07
Improved prepended/appended test

smithdc1:
Changed prepended/appended test to assert against the whole form output rather than code snippets.


smithdc1:
Need to think about how to test this part of `PrependedAppendedText` more. 

``` python
        # Bootstrap 3
        if "input-lg" in css_class:
            self.input_size = "input-lg"
        if "input-sm" in css_class:
            self.input_size = "input-sm"
```

PR #1255, opened 2022-05-17
Completed improved doc string and __init__ signatures for remaining layout objects

smithdc1:


smithdc1:
Side note @carltongibson 

https://github.com/django-crispy-forms/django-crispy-forms/runs/6478218649?check_suite_focus=true#step:7:152

One test failure here with Django 4.1. It's due to the cached template loader now being enabled in development. 
https://github.com/django/django/blob/2cec020f5bc462954d902bdad1a5955852cecff6/docs/releases/4.1.txt#L549

PR #1256, opened 2022-05-20
Expanded example in crispy_tag_formsets.rst

bill-ash:
Hello - just a small change that caused me quite a bit of pain when reading. 

Hope it helps others!

### DIFF
--- a/docs/crispy_tag_formsets.rst
+++ b/docs/crispy_tag_formsets.rst
@@ -65,8 +65,18 @@ Extra context
 
 Rendering any kind of formset with crispy injects some extra context in the layout rendering so that you can do things like::
 
-    HTML("{% if forloop.first %}Message displayed only in the first form of a formset forms list{% endif %}",
-    Fieldset("Item {{ forloop.counter }}", 'field-1', [...])
+    class ExampleFormSetHelper(FormHelper): 
+        def __init__(self, *args, **kwargs): 
+            super(FormHelper, self).__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


docs/crispy_tag_formsets.rst
```diff
@@ -65,8 +65,18 @@ Extra context
...
+        def __init__(self, *args, **kwargs): 
+            super(FormHelper, self).__init__(*args, **kwargs)
+            self.form_method = 'post'
+            self.layout = Layout(
+                HTML('{% if forloop.first %} Only display text on the first iteration... {% endif %}'),
+                Fieldset('Item: {{forloop.counter}}', 'field'),
+                'favorite_color', 
+                'favorite_food',
+            )
+            self.render_requried_fields = False
```
smithdc1, line 78: Do we need to set this for a minimum example? 🤔

smithdc1 (commented):
Thanks. Looks good, just one small question.

docs/crispy_tag_formsets.rst
bill-ash: Thought it was nice to display control of the formset from the helper attributes. But, not needed for a small example.

smithdc1: Understood, in that case could we remove it for here. We could show `render_required_fields` but I think that should be a standalone example. 

Thanks again for the PR. 🥇

smithdc1, line 78:
```suggestion
```

PR #1262, opened 2022-06-21
Used optgroups filter to render checkbox and radio in Bootstrap3 template pack. 

smithdc1:
Alternative to #1199


smithdc1:
@nijel does this work for you?

PR #1276, opened 2022-06-27
Removed `|safe` from fieldset template.

smithdc1:


crispy_forms/layout.py
```diff
@@ -560,9 +560,10 @@ def __init__(self, legend, *fields, css_class=None, css_id=None, template=None,
     def render(self, form, context, template_pack=TEMPLATE_PACK, **kwargs):
         fields = self.get_rendered_fields(form, context, template_pack, **kwargs)
 
-        legend = ""
         if self.legend:
-            legend = "%s" % Template(str(self.legend)).render(context)
+            legend = Template(str(self.legend)).render(context)
```
smithdc1, line 564: `Template.render()` returns a `safestring`, so let's not convert it back to a `str`

PR #1290, opened 2022-07-06
Refactored FormHelper.get_attributes() to create attrs dict first.

smithdc1:


crispy_forms/helper.py
```diff
@@ -324,20 +336,6 @@ def get_attributes(self, template_pack=TEMPLATE_PACK):  # noqa: C901
                 offset_pattern = "col-%s-offset-%s"
                 items["bootstrap_checkbox_offsets"] = [offset_pattern % m for m in bootstrap_size_match]
 
-        items["attrs"] = {}
-        if self.attrs:
-            items["attrs"] = self.attrs.copy()
-        if self.form_action:
-            items["attrs"]["action"] = self.form_action.strip()
```
smithdc1, line 331: Mypy does't like creating a dict in this way.

We can simplify this by creating this attrs dict first then adding them to items above.

It's somewhat a shame that `flatatt` keeps items where the value is `""`. This means we need to strip those items out here. 

Alternative would be to change default values to `None` and drop the `strip()` but the seems like some churn we can avoid at this stage.

PR #1292, opened 2022-07-07
Update docs 'FormHelper' examples to use 'property'

pawelad:
I found that there is a mismatch between what the docs say, and what the `FormHelper` docstring says:

https://github.com/django-crispy-forms/django-crispy-forms/blob/6477527efbf9943140478595d2ef6211d56ed8ae/crispy_forms/helper.py#L161-L174

I like the `property` way much more then the overriding `__init__` and wanted to unify the recommendation across both places.

Let me know what you think!

### DIFF
--- a/docs/crispy_tag_forms.rst
+++ b/docs/crispy_tag_forms.rst
@@ -52,10 +52,12 @@ Your helper can be a class level variable or an instance level variable, if you 
     from crispy_forms.helper import FormHelper
 
     class ExampleForm(forms.Form):
-        [...]
-        def __init__(self, *args, **kwargs):
-            super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


smithdc1:
This is interesting I've never really thought about using it in this way. 


Let me have a think. 

Initial thought is maybe it needs a few words as an intro. Something about it needing to be an attribute of the form and that a nice way to do this is with a property.

pawelad:
I think I closed this PR by accident when cleaning up my GitHub forks - sorry!

Would you like to update the docs to suggest the `@property` way (or mention both ways)? If not, should we update the `FormHelper` docstring?  I can take a stab at either one of those, let me know.

PR #1305, opened 2022-11-09
Update setup-python and use its builtin cache feature

spenserblack:


smithdc1:
Thanks for the contribution 👍

PR #1313, opened 2022-12-17
Dropped bootstrap4 template pack.

smithdc1:
There's also a default for `ALLOWED_TEMPLATE_PACKS`. I'm not sure we need to go as far to remove bootstrap4 here. That means everyone using bootstrap4 would have to add an additional setting. If we don't change it, the upgrade path should just be install the template pack and add it to your list of installed apps. 

https://github.com/django-crispy-forms/django-crispy-forms/blob/d240bacc1a7f5797531c55178cc764df55fcda9f/crispy_forms/templatetags/crispy_forms_tags.py#L260


requirements/testing.txt
```diff
@@ -2,3 +2,4 @@ coverage
 pytest
 pytest-cov
 pytest-django
+git+https://github.com/django-crispy-forms/crispy-bootstrap4
```
smithdc1, line 5: We need multiple template packs available to test the overriding of template packs.

PR #1323, opened 2022-12-31
Moved coverage config to pyproject.toml

smithdc1:
toml was added to the standard library in Python 3.11. Therefore need to install the toml extra with coverage for Python3.7-3.10. Refs https://coverage.readthedocs.io/en/7.0.1/config.html#configuration-reference

### DIFF
--- a/requirements/testing.txt
+++ b/requirements/testing.txt
@@ -1,4 +1,4 @@
-coverage
+coverage[toml]
 pytest
 pytest-cov
 pytest-django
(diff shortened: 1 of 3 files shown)


requirements/testing.txt
```diff
@@ -1,4 +1,4 @@
-coverage
+coverage[toml]
```
smithdc1, line 1: Is this how we install extras? Let's see what the test suite says. It seems to work locally, but that's with Python 3.11...

smithdc1: https://github.com/django-crispy-forms/django-crispy-forms/actions/runs/3813140645/jobs/6486715584#step:6:42

Coverage continues to work all the way back to Python 3.7, with branch reporting.

PR #1328, opened 2023-01-18
Added Django 4.2 support.

smithdc1:


smithdc1:
Ah. 

The package publish step failed with  `Error: Classifier 'Framework :: Django :: 4.2' is not a valid classifier.`

Refs: https://github.com/pypa/trove-classifiers/pull/128

PR #1332, opened 2023-02-02
Drop Django 4.2 Classifier.

smithdc1:
The 4.2 classifier is not yet available and is blocking the 2.0a1 release.


smithdc1:
Tests failing as new version of Black has been released. 🙄

Issue #1336, opened 2023-02-15
Crispy forms templates won't load for django-crispy-forms 2.0


mikbuch:
* Package version: 2.0
* Django version: 4.1.4 & 4.1.7 & 3
* Python version: 3.10
* Template pack: bootstrap4, but it affects all template packs

### Description:

Crispy forms templates won't load on template rendering.

Solution: downgrade crispy forms to 1.14.0

Probable cause: crispy_forms package that is installed to site-packages is missing "templates" directory? Very strange.

### Preferably also include:

- [x] Example Django Crispy Forms code

```
{% load crispy_forms_tags %}

        <div class="login-page">
            <form class="login-container" method="POST">
                {% csrf_token %}
                {{ login_form|crispy }}
                <div class="btn-action-wrapper">
                    <button class="btn btn-primary" type="submit" style="margin-top: 10px">Login</button>
                </div>
            </form>
        </div>
```

- [x] Screenshots

<img width="2447" alt="image" src="https://user-images.githubusercontent.com/10733514/219179095-0ad97e48-5523-4276-8f02-4b8e915bd487.png">

- [x] Actual HTML generated

None, error, either Django debug or 500 server error.

- [x] Expected HTML

```
<html><head></head><body><div class="login-page">
            <form class="login-container" method="POST">
                <input type="hidden" name="csrfmiddlewaretoken" value="vY6tZVikuKEA1fPDkmfIQ01MeOqrGZMO7CCJDfnMYJjnCX8uAzeMVMLoOFNpfF66">
                

    






    


    <div id="div_id_username" class="ctrlHolder">
        
            
        

        
            

            <label for="id_username" class="requiredField">
                Username<span class="asteriskField">*</span>
            </label>
        

        
            <input type="text" name="username" autofocus="" autocapitalize="none" autocomplete="username" maxlength="150" class="textinput textInput" required="" id="id_username">
        

        
    </div>



    


    <div id="div_id_password" class="ctrlHolder">
        
            
        

        
            

            <label for="id_password" class="requiredField">
                Password<span class="asteriskField">*</span>
            </label>
        

        
            <input type="password" name="password" autocomplete="current-password" class="textinput textInput" required="" id="id_password">
        

        
    </div>






                <div class="btn-action-wrapper">
                    <button class="btn btn-primary" type="submit" style="margin-top: 10px">Login</button>
                </div>
            </form>
        </div>
</body></html>
```

mikbuch:
[...]

PR #1338, opened 2023-02-20
Template pack example install added to the docs (#1336)

mikbuch:
Also minor reformatting of this docs page.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -34,14 +34,19 @@ In production environments, always activate Django template cache loader. This i
 
 Template packs
 ~~~~~~~~~~~~~~
-Since version 2.0, django-crispy-forms has built-in support for version 3 and 4 of the Bootstrap CSS framework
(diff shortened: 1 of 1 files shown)


smithdc1:
Fixed in 443ebd208d0945b87e063c6d780e93ba599da17a

mikbuch:
@smithdc1 , thank you for modifying the docs. My only remark is that in my commit I explicitly added an example install line:

```
pip install crispy-bootstrap4
```

see: 

https://github.com/django-crispy-forms/django-crispy-forms/pull/1338/commits/108ef405422c3d37bd72ce7f97319c768c7d1514

In this type of "getting started" pages people are used to just copy-paste line after the line, hoping to get things working, at least for the default settings. This was the case for the pre `2.0` `django-crispy-forms`. I understand that the links are there, and the people can click the link and find a wasy/a command to install what is needed, but the thing is that 90% of people visiting this getting started page are in a hurry, or on some kind of a deadline, so they would rather have everything in one place. I am afraid that rather than looking trough the links and reading through the documentation they would rather go to a different template manager/pack than `django-crispy-forms`, which would be a shame, because `crispy-forms` are great!

PR #1340, opened 2023-02-24
Replace 'media' with 'static'

DanielSwain:
The word 'static' is used in the description in one place.  This change fixes the wording in the header and also in one other place in the description.


smithdc1:
Super! Thank you!

smithdc1 (approved):

PR #1341, opened 2023-03-01
Submit, Reset btn default css_class changed

Shekhrozx:
(from this stackoverflow issue: https://stackoverflow.com/q/75603468/7456750)

There is a default CSS class like `btn-primary` and `btn-inverse` in the `Submit` and `Reset` (input) buttons.
```py
class Submit(BaseInput):
    # ...
    input_type = "submit"
    field_classes = "btn btn-primary"
```
```py
class Reset(BaseInput):
    # ...
    input_type = "reset"
    field_classes = "btn btn-inverse"
```

So the default CSS class is changed. Although when users add new CSS classes to `css_classes` they are added after the default classes. So let users modify the `Submit` and `Reset` btn classes.

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -321,7 +321,7 @@ class Submit(BaseInput):
     """
 
     input_type = "submit"
-    field_classes = "btn btn-primary"
+    field_classes = "btn"
 
 
 class Button(BaseInput):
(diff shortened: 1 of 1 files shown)


smithdc1:
Thanks for the patch. 

I think for a default I think `btn-primary` for `Submit` and `btn-inverse` for `Reset` are sensible values to start with. I also appreciate that some folk may want to change the default values. I'd be happy to accommodate such a change but only if we can make it in a way that's backward compatible. As it is this is a breaking change with _many, many_ folk will be relying on this as it's been like this _~forever_. 

My immediate reaction is that I think having a small subclass as in [this answer](https://stackoverflow.com/a/75604240/12482183) on stackoverflow is the way to go.

PR #1342, opened 2023-03-02
Updated install docs to list new template packages

sankarebarri:
I came across unusuall problem after installing crispy-forms v2.0. After some googling, i fell on a solution on StackOverFlow and worried why it was not mentioned on the crispy-forms official documentation site.

This is the link to the StackOverFlow solution https://stackoverflow.com/questions/75495403/django-returns-templatedoesnotexist-when-using-crispy-forms


smithdc1:
Thanks @sankarebarri for the patch. 👍

PR #1354, opened 2023-06-15
Test against 3.12-dev & add Python 3.12 classifier

foarsitter:
Ran the tests against 3.12-dev and they succeed. Good job everybody! 

I didn't added 3.12 to the publish version matrix since 3.12 better may not block an upcoming release.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,6 +19,7 @@ jobs:
         - '3.9'
         - '3.10'
         - '3.11'
+        - '3.12-dev'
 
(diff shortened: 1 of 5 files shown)


foarsitter:
> 
> I think let's add it. Better do it now than forget later.

Alright, pushed.

smithdc1:
Great. 👍

Could you add a small note to the changelog?

foarsitter:
Of course, done!

smithdc1:
Thank you for the contributions!

PR #1358, opened 2023-07-03
Replace trunk with main/upstream in contributing.rst

ckrybus:
Seeing "trunk" in contributing.rst is confusing so I removed it. It turns out the instructions are quite old, from 2012, probably when the project was still hosted on code.google.com.


smithdc1:
Thank you! 🎁

PR #1363, opened 2023-08-11
Fixes #1361 -- Improved Fieldset types.

### LINKED ISSUE #1361: Allow Fieldset to accept Fields
* Package version: 2.0
* Django version: 3.2.18
* Python version: 3.8.17
* Template pack: crispy-bootstrap5 0.7

### Description:

The `Fieldset` class accepts a number of fields to render in a `<fieldset>`. Apparently, the value of this field can only be `str`:

https://github.com/django-crispy-forms/django-crispy-forms/blob/3b81cc98f99505cd55a7fdc7cb7ca2744f6e2546/crispy_forms/layout.py#L521

This causes a type checking error if one uses `Field` and any subclasses of it, like `AppendedText` from the Bootstrap pack.

I propose changing the type of the kwargs to `str | LayoutObject`

smithdc1:
Refs #1361
@kytta -- Does this work for you?

### DIFF
--- a/crispy_forms/layout.py
+++ b/crispy_forms/layout.py
@@ -518,7 +518,7 @@ class Fieldset(LayoutObject):
     legend : str
         The content of the fieldset's ``<legend>``. This text is context
         aware, to bring this to life see the examples section.
-    *fields : str
+    *fields : str | LayoutObject | HTML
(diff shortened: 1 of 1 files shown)


kytta (approved):
Looks good, thanks!

PR #1375, opened 2024-01-08
Fix documentation of `Field`'s constructor

bblanchon:
The documentation of `kwargs` for `Field`'s constructor was incorrect.
It said that the attributes are added to the `<div>`, but they are actually added to the `<input>`.


smithdc1 (approved):
Thanks for the patch!

PR #1377, opened 2024-02-07
added context to Fieldset render

fwannmacher:
Adding the context to the Fieldset render allows its variables to be accessed in custom templates.


fwannmacher:
Any idea on when this would be merged @smithdc1 ?

smithdc1:
Hi @fwannmacher I want to sit down with this and have a play. Does it break anything (its been like this _forever_), what about docs etc. I need a chunk of time to do that. 

It's on the list, but I won't get to it right now (it's currently holiday time here).

fwannmacher:
Hi @smithdc1 I'm working in project where it is need unfortunately. We need to have access to the context in the fieldsets so we can do some stuff in its custom templates. I guess either few people use fieldsets or few people use custom templates :)

PR #1391, opened 2024-06-11
Revert "Allowed pass through of context when rendering a layout object. (#1377)"

smithdc1:
This reverts commit f9d67d3e164c52f05a3465bce6aada099ad6fd9f.

I'm a bit confused to be honest. I am pretty certain I saw the checks pass on the PR page after I pushed my edits, but clearly that wasn't the case. 

The error is a bit odd -- the tests don't even run! Instead the error message is _"GitHub Actions has encountered an internal error when running your job"_. Anyway let's see if reverting this fixes the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## 2.2 (TBC)
 * Added support for Django 5.1.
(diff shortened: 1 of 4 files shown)


smithdc1:
Seems github actions must have had some downtime. Tests are green on main now.

PR #1398, opened 2024-07-25
Merge fix

kosdmit:
We lost some changes from the main in this branch. I moved the render method from Container to Tab (to fix the issue with `active` css class in accordion-group), and added types to it.

### DIFF
--- a/crispy_forms/bootstrap.py
+++ b/crispy_forms/bootstrap.py
@@ -799,7 +799,13 @@ class Tab(Container):
         link_template = self.link_template % template_pack
         return render_to_string(link_template, {"link": self})
 
-    def render(self, form, context, template_pack=TEMPLATE_PACK, **kwargs):
+    def render(
+        self,
+        form: BaseForm,
+        context: Context,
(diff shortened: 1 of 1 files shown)


smithdc1:
Thank you for looking at this. 


I'm on holiday at the moment so can't look at this properly right now but just on this point :

> but I still have mypy errors with Python 3.8

Given Python 3.8 support ends in ~3 months I'd consider dropping Python 3.8 support if it's a lot of extra work to maintain compatibility for this version. Maybe by the time we get it reviewed/merged/released we'd be at that point anyway.

kosdmit:
> I'd consider dropping Python 3.8 support if it's a lot of extra work to maintain compatibility for this version.

There are just a few extra changes to keep python 3.8 compatibility. 

> Mypy runs in all CI jobs, because i get different behavior depending on the python version. For example, the python 3.9 version passes successfully, but I still have mypy errors with Python 3.8.

I talked there about reason for running mypy on all supported python versions. Original CI workflow ran mypy only on python 3.9 CI job.

PR #1405, opened 2024-11-26
Allow tabs other than the first one to be set active

gilbertf:
Currently the active properties set in the function signatures Tab(...,active=None,...) are reset to None by the TabHolder render() function. Instead the first Tab is set active (in case no Tab has Errors).

With this change making the second tab active is now working as expected:

`TabHolder(
Tab("First", active=False),
Tab("Second", active=True)`


smithdc1:
Thanks for the patch. Please can you look at the failing test(s) as we should avoid a regression with this patch. A new test to show how your proposed fix works would also be very helpful.

PR #1413, opened 2025-04-13
PEP 639 licenses 

smithdc1:
Warnings are now being raised when building the package related to this change. 

https://peps.python.org/pep-0639/ specifies how licenses are to be documented.

The guide here https://packaging.python.org/en/latest/guides/licensing-examples-and-user-scenarios/ is helpful.

1. Switch the 'MIT License' classifier to use 'license = "MIT"' instead. (As in the 'basic example' in the PyPA guide.
2. Add a 'license-files' for the license where we're using a non-standard (?) 'LICENSE.txt' file name.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,6 @@ classifiers=[
     "Framework :: Django :: 5.0",
     "Framework :: Django :: 5.1",
     "Framework :: Django :: 5.2",
-    "License :: OSI Approved :: MIT License",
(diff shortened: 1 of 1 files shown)


smithdc1:
- Changes to the licenses need to be made before Feb-2026.
- And requires a recent version of setuptools which only supports Pyhon 3.9+.

Therefore, we should merge this after we drop Python 3.8 (sometime after haloween this year).

PR #1416, opened 2025-05-26
Enabled GitHub code of conduct.

smithdc1:
Adding a reference to say that we follow Django's code of conduct.

See [docs](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-code-of-conduct-to-your-project) and discussion from when this was added to Django itself: https://github.com/django/django/pull/15057


smithdc1:
Once merged here I'll add to the other projects in the organisation.

PR #1418, opened 2025-06-20
Don't crash when using the crispy template tag and missing variables

bmispelon:


tests/test_tags.py
```diff
@@ -216,3 +216,15 @@ def test_crispy_addon():
...
         crispy_addon(bound_field)
+
+
+def test_crispy_missing_variable():
+    template = Template("{% load crispy_forms_tags %}{% crispy form %}")
+    html = template.render(Context({}))
+    assert html == ""
+
+
+def test_crispy_missing_variable_deep():
```
bmispelon, line 227: I wanted to have a second test with nested variables where the first resolves correctly but the second doesn't.

PR #1434, opened 2026-04-25
docs: add multifield.html to template pack documentation

### LINKED ISSUE #1393: The multifield template is missed in docs.template_packs
* Package version: Any 
* Django version: Any
* Python version: Any

### Description:

I have noticed that we are have the same issue (missed multifield.html) in different template packs: 

- [crispy-bootstrap5 issue #75](https://github.com/django-crispy-forms/crispy-bootstrap5/issues/75)
- issue #1056
- issue #478

Probably it is going because of we have legacy example for creating new template packs in docs. This part of docs gives list of templates which does not includes multifield.html

- [ ] Screenshots
image

Thiago88pe:
Adds mention of multifield.html to the template pack docs. Fixes #1393.

### DIFF
--- a/docs/template_packs.rst
+++ b/docs/template_packs.rst
@@ -56,7 +56,7 @@ If we are planning on using formsets + ``{% crispy %}`` we will need a ``whole_u
 
 All of these templates use a tag named ``{% crispy_field %}`` that is loaded doing ``{% load crispy_forms_field %}``, that generates the html for ``<input>`` using ``field.html`` template, but previously doing Python preparation beforehand. In case you wonder the code for this tag lives in ``crispy_forms.templatetags.crispy_forms_field``, together with some other stuff.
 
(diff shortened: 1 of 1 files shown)


smithdc1:
Any thoughts on the file tree? The issue says the template is not listed in the docs? 🤔

> This part of docs gives list of templates which does not include multifield.html

Thiago88pe:
Hi @smithdc1, thanks for the review!

Good point — I should have explained the reasoning in the PR description. Looking at the file tree on `main`, `layout/multifield.html` is already listed there, so the tree itself looks fine.

What I noticed is that the prose right below the tree (in the *Fundamentals* section) walks through which templates you need for specific layout objects — `div.html` for `Div`, `baseinput.html` for `Submit`, etc. — but `MultiField` / `multifield.html` wasn't mentioned. My change adds that one sentence so the explanation is consistent with the other layout objects.

That said, if you read #1393 as strictly about the tree (which is a fair reading), then the issue may already be resolved on `main` and this PR is more of a small docs consistency improvement. Happy to reframe the PR description that way, or to close it if you don't think it's worth merging — just let me know what you prefer.

PR #1439, opened 2026-08-02
Add `formset.non_form_errors` to formset example

MrQubo:
Formsets not rendered directly via `{% cripsy formset %}` should use `formset.non_form_errors`. Otherwise these errors won't show up to the user.


MrQubo:
This was a simple change and I didn't bother reading contribution guide. Sorry in advance if I did something wrong. :sweat_smile:

